PR #9, opened 2017-12-17
EEI: properly handle out of gas exceptions

axic:


src/hera.cpp
```diff
@@ -107,19 +107,26 @@ EXPORT struct evm_result evm_execute(struct evm_instance* instance,
...
-  // copy call result
-  ret.output_size = call->returnValue.size();
-  ret.output_data = (const uint8_t *)malloc(ret.output_size);
-  // FIXME: properly handle memory allocation issues
-  if (ret.output_data) {
-    std::copy(call->returnValue.begin(), call->returnValue.end(), (char *)ret.output_data);
+  if (ret.gas_left != EVM_EXCEPTION) {
+    // copy call result
+    ret.output_size = call->returnValue.size();
+    ret.output_data = (const uint8_t *)malloc(ret.output_size);
```
jakelang: use calloc() instead for safety; uninitialized memory can cause some nasty problems.

axic: True, but the entire buffer is overridden below. This assumes `std::copy` is working as intended.

PR #10, opened 2017-12-17
External Interface: Helper functions

jakelang:
-Helper function to switch endianness of a given byte array.
-Simpler memory access functions

### DIFF
--- a/src/eei.h
+++ b/src/eei.h
@@ -91,6 +91,20 @@ private:
      }
   }
 
+  void endianSwp(uint8_t *bytes, size_t len)
+  {
+    int i = 0;
+    int j = len - 1;
+
+    while (i < j) { /* XOR swap each byte of array */
+      bytes[i] ^= bytes[j];
+      bytes[j] ^= bytes[i];
+      bytes[i] ^= bytes[j];
+
+      ++i;
+      --j;
+    }
+  }
 
 private:
   Hera *hera;
--- a/src/eei.h
+++ b/src/eei.h
@@ -75,6 +75,23 @@ private:
     }
   }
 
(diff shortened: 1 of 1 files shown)


src/eei.h
```diff
@@ -75,6 +75,37 @@ struct EthereumInterface : ShellExternalInterface {
     }
   }
 
+  void memRead(uint32_t offset, uint8_t *dst, size_t length)
+  {
+     uint32_t i;
+     for (i = offset; i < (offset + length); i++) {
```
axic: Type declaration could be folded into the for loop, though this hear is susceptible to overflow (size requirement of offset + length might be bigger than the size of i)

src/eei.h
```diff
@@ -75,6 +75,37 @@ struct EthereumInterface : ShellExternalInterface {
...
+
+  void memWrite(uint32_t dstoffset, uint8_t *src, size_t length)
+  {
+     uint32_t i;
+     for (i = dstoffset; i < (dstoffset + length); i++) {
+     	memory.set<uint8_t>(i, *(src + i - dstoffset));
+     }
+  }
+
+  void endianSwp(uint8_t *bytes, size_t len)
```
axic: I prefer non-shortened names, such as `endianSwap`.

src/eei.h
```diff
@@ -75,6 +75,37 @@ struct EthereumInterface : ShellExternalInterface {
...
+  {
+    int i = 0;
+    int j = len - 1;
+
```
axic: Needs an assert that `len` is even.

jakelang: This xor reverse algorithm should accommodate all non-zero values of len, no?

Issue #3, opened 2017-12-17
Update to latest EVM-C / EEI


axic:
The latest version of the EVM-C spec can be found here: https://github.com/ethereum/evmjit/blob/develop/include/evm.h

It has 4 data types:
- arbitrary memory buffer
- big endian 256 bit number
- big endian 160 bit number (address)
- host endian 64 bit number (gas, etc.)

The latest version of the EEI spec can be found here: https://github.com/ewasm/design/blob/master/eth_interface.md#data-types

It has 4 data types:
- arbitrary memory buffer
- little endian 256 bit number
- little endian 160 bit number (address)
- little endian 128 bit number (value)

axic:
I think we need helpers (the name implies endian translation to/from the host endianness)
- `loadMemory`
- `storeMemory`
- `readUint256`
- `readUint160`
- `readUint128`
- `exceedsUint128` (not sure about naming, a check that the host supplied 256-bit number doesn't have a value exceeding 128 bits)

axic:
Also realised that `memoryCopy` can be replaced with `storeMemory`.

PR #11, opened 2017-12-30
EEI: synchronous methods

jakelang:
#3

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -137,7 +137,7 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       heraAssert((srcOffset + length) > srcOffset, "Out of bounds (source) memory copy.");
 
       for (uint32_t i = 0; i < length; ++i) {
-          *(dst + i) = memory.get<uint8_t>(srcOffset + i);
+          dst[length - (i + 1)] = memory.get<uint8_t>(srcOffset + i);
       }
   }
 
@@ -148,7 +148,7 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       heraAssert(memory.size() < (dstOffset + length), "Out of bounds (destination) memory copy.");
 
       for (uint32_t i = 0; i < length; ++i) {
-          memory.set<uint8_t>(dstOffset + i, *(src + i));
+          memory.set<uint8_t>(dstOffset + length - (i + 1), *(src + i));
       }
   }
 
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -56,7 +56,7 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
 
       cout << resultOffset << "\n";
 
-      copyAddressToMemory(msg.address, resultOffset);
+      storeUint160(resultOffset, msg.address);
 
       return Literal();
     }
@@ -124,14 +124,6 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
 
     for (; i < (srcoffset + length); i++, j++) {
       memory.set<uint8_t>(j, src[i]);
-    }
-  }
-
-  void EthereumInterface::copyAddressToMemory(struct evm_address const& address, uint32_t dstoffset)
-  {
-    heraAssert(memory.size() < (dstoffset + 20), "Out of bounds (destination) memory copy.");
-    for (uint32_t i = 0, j = dstoffset; j < (dstoffset + 20); i++, j++) {
-      memory.set<uint8_t>(j, address.bytes[i]);
     }
   }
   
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -57,6 +57,83 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      int64_t gas = arguments[0].geti64();
+      uint32_t addressOffset = arguments[1].geti32();
+      uint32_t valueOffset = arguments[2].geti32();
+      uint32_t dataOffset = arguments[3].geti32();
+      int32_t dataLength = arguments[4].geti32();
+      uint32_t resultOffset = arguments[5].geti32();
+      int32_t resultLength = arguments[6].geti32();
+
+      evm_result *call_result = new evm_result;
+      evm_message *call_message = new evm_message;
```
axic: I think we should have a helper for building a `call_message`, it is going to be shared by all the different call implementations.

jakelang: this is a good idea, since callCode and other interface methods pretty much repeat all of this

src/eei.h
```diff
@@ -79,6 +79,109 @@ struct EthereumInterface : ShellExternalInterface {
...
+      for (uint32_t i = 0; i < length; ++i) {
+          memory.set<uint8_t>(dstOffset + i, *(src + i));
+      }
+  }
+
+  /*
+   * Memory Op Wrapper Functions
+   */
+
+  void loadUint256(uint32_t srcOffset, struct evm_uint256be *dst)
```
chfast: Just return `evm_uint256be`. Similarly in other helpers.

src/eei.h
```diff
@@ -79,6 +79,109 @@ struct EthereumInterface : ShellExternalInterface {
...
+      endianSwap(src->bytes, 32);
+      storeMemory(dstOffset, src->bytes, 16);
+  }
+
+  /*
+   * Utilities
+   */
+  
+  /* Checks if host supplied 256 bit big-endian value exceeds UINT128_MAX */
+  unsigned int exceedsUint128(const struct evm_uint256be *value)
```
chfast: In general, in C++ (otherwise than in C) you can skip the `struct` keyword. Also for struct arguments pass them by reference `const evm_uint256be& value`. You will not have to check if the `value != nullptr`.

src/eei.cpp
```diff
@@ -57,6 +57,83 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+    if (import->base == Name("getBalance")) {
+      std::cout << "getbalance\n";
+
+      uint32_t addressOffset = arguments[0].geti32();
+      uint32_t resultOffset = arguments[1].geti32();
+
+      evm_address *address = new evm_address;
+      evm_uint256be *result = new evm_uint256be;
+
+      loadUint160(addressOffset, address);
```
chfast: No need for new/delete, just place them on the stack. With other changes applied it should simplify to

```
evm_address address = loadUint160(addressOffset);
context.fn_table->get_balance(result, &context, &address);
```

src/eei.h
jakelang: The helper functions already have these changes in #19 .

src/eei.h
jakelang: done on the helper functions branch now.

src/eei.cpp
```diff
@@ -48,15 +48,85 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      int32_t vm_status = 0;
+      uint8_t *input_data = new uint8_t[dataLength];
+
+      call_message.address = loadUint160(addressOffset);
+      call_message.sender = msg.address;
+      call_message.value = loadUint128(valueOffset);
+      loadMemory(dataOffset, input_data, dataLength);
+      call_message.input = input_data;
+      call_message.input_size = dataLength;
+      std::memset(&call_message.code_hash, 0, sizeof(evm_uint256be));
```
chfast: The easier way to do it is `call_message.code_hash = {}` or `call_message.code_hash = {{0,}}`.

src/eei.cpp
```diff
@@ -48,15 +48,85 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      uint32_t valueOffset = arguments[2].geti32();
+      uint32_t dataOffset = arguments[3].geti32();
+      int32_t dataLength = arguments[4].geti32();
+      uint32_t resultOffset = arguments[5].geti32();
+      int32_t resultLength = arguments[6].geti32();
+
+      evm_result *call_result;
+      evm_message call_message;
+      int32_t vm_status = 0;
+      uint8_t *input_data = new uint8_t[dataLength];
```
chfast: You should not use `new` / `delete` directly. Here you can use `std::vector input_data(dataLength)`.

But it might be better to actually `loadMemory` to return `std::vector` with the memory copy.

chfast:
You can try it like this:
```
void loadMemory(char* out_data, size_t size);

std::vector<char> loadMemoryBuffer(size_t size)
{
    std::vector<char> buf(size);
    loadMemory(buf.data(), size);
    return buf;
}

```

This is not ideal, because the vector will be filled with zeros first (unnecessarily).

PR #13, opened 2018-01-09
Reworked helper functions

jakelang:
#3 
New: integrated endianness switching, wrappers for helper functions, and possibly redundant null pointer checks.
 
note: obsoletes #10


src/eei.h
```diff
@@ -75,6 +75,112 @@ struct EthereumInterface : ShellExternalInterface {
...
+      endianSwap(src->bytes, 16);
+      storeMemory(dstOffset, src->bytes, 16);
+  }
+
+  /*
+   * Utilities
+   */
+  
+  /* Checks if host supplied 256 bit value exceeds UINT128_MAX */
+  int exceedsUint128(struct evm_uint256be *value)
```
axic: Can you please use `unsigned` here?

src/eei.h
```diff
@@ -75,6 +75,112 @@ struct EthereumInterface : ShellExternalInterface {
...
+  {
+      int doesExceed = 0;
+
+      assert(value != nullptr);
+      for (int i = 0; i < 16; ++i) {
+        if (value->bytes[i])
+	    doesExceed = 1;
```
axic: Can you please do `return 1` here and `return 0` outside the loop otherwise?

PR #17, opened 2018-01-09
[WIP] Run stEWASMTests with testeth

jwasinger:
Depends on https://github.com/ethereum/cpp-ethereum/pull/4764, https://github.com/ethereum/tests/pull/391 and https://github.com/ethereum/tests/pull/394

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -43,6 +43,19 @@ defaults:
       name: "Restore dependencies cache"
       key: *deps-cache-key
 
+  test-ewasm: &test-ewasm
+    run:
+      name: "Run ewasm state tests"
+      working_directory: ~/project
+      command: |
+        echo "print working directory:"
+        pwd
+        echo "list files in working directory:"
+        ls -al .
+        echo "run tests script."
+        ./scripts/tests.sh
+
+
   linux-steps: &linux-steps
     - checkout
     - *update-submodules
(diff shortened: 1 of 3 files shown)


axic:
```
$ ./scripts/build.sh
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

You need to add the `sudo: true` option to travis.

axic:
Need to add ccache otherwise this will be prohibitively slow.

axic:
```
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```

Fails with the above but has `sudo: required` in travis, @chfast do you have any idea?

scripts/tests.sh
```diff
@@ -0,0 +1,18 @@
...
+ln -s `pwd`/../. hera
+
+mkdir build
+cd build
+cmake -DHERA=ON ..
+make
+
+git clone --recursive https://github.com/ewasm/tests -b wasm-tests
+
+ETHEREUM_TEST_PATH='`pwd`/tests`' test/testeth -t GeneralStateTests/stEWASMTests -- --filltests --vm hera --singlenet "Byzantium"
```
axic: This shouldnt be filltest but rather running the tests.

scripts/tests.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+git clone --recursive https://github.com/ethereum/cpp-ethereum
+cd cpp-ethereum
+
+rm -rf hera
+ln -s `pwd`/../. hera
```
axic: Should make this relative to dirname of the current script

axic:
This needs circleci integration.

axic:
@jwasinger can you try the updated script?

cdetrio:
CircleCI (circle.yaml) was added here https://github.com/ewasm/hera/pull/94

axic:
The first step in this PR is to make sure `scripts/test.sh` works locally. After that, a proper cirecle target must be added.

cdetrio:
@axic @jwasinger i think this is ready now

circle.yml
```diff
@@ -43,6 +43,19 @@ defaults:
...
+    run:
+      name: "Run ewasm state tests"
+      working_directory: ~/project
+      command: |
+        echo "print working directory:"
+        pwd
+        echo "list files in working directory:"
+        ls -al .
+        echo "run tests script."
+        ./scripts/tests.sh
```
jakelang: Seeing as this script clones and rebuilds cpp-ethereum each time, it will be very slow to run tests without using the dependency cache. I would suggest moving the client build step out of the script and into a separate CI job where we can add the caching.

jakelang: UPDATE:
tried doing this. I think it would be too messy without setting up a CMake dependency for cpp-ethereum, or a submodule.

PR #19, opened 2018-01-10
Reworked helper fns

### LINKED ISSUE #3: Update to latest EVM-C / EEI

jakelang:
Fixes #3 - helper functions for interface methods
also cleaned up the commit history

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -56,7 +56,7 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
 
       cout << resultOffset << "\n";
 
-      copyAddressToMemory(msg.address, resultOffset);
+      storeUint160(msg.address, resultOffset);
 
       return Literal();
     }
@@ -75,8 +75,8 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
 
       cout << resultOffset << " " << dataOffset << " " << length << "\n";
 
-      vector<char> input(msg.input, msg.input + msg.input_size);
-      memoryCopy(input, dataOffset, resultOffset, length);
+      vector<uint8_t> input(msg.input, msg.input + msg.input_size);
+      storeMemory(input, dataOffset, resultOffset, length);
 
       return Literal();
     }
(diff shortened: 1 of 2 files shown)


src/eei.h
```diff
@@ -64,7 +64,20 @@ struct EthereumInterface : ShellExternalInterface {
   void memoryCopy(std::vector<char> const& src, uint32_t srcoffset, uint32_t dstoffset, uint32_t length);
   void copyAddressToMemory(struct evm_address const& address, uint32_t dstoffset);
```
axic: I think this s the same as `storeUint160`. We should remove this one.

src/eei.cpp
```diff
@@ -132,4 +131,108 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+        if (value->bytes[i])
+	    return 1;
+      }
+      return 0;
+  }
+
+  /* Endianness Converter */
+  void EthereumInterface::endianSwap(uint8_t *bytes, size_t len)
+  {
+  	assert(len);
```
axic: Can you make this more explicit with `len > 0`?

axic:
Please use `heraAssert`s in a similar manner as in `memoryCopy` and `loadAddressFromMemory`.

src/eei.cpp
```diff
@@ -119,12 +118,108 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      for (uint32_t i = 0; i < length; ++i) {
+          memory.set<uint8_t>(dstOffset + i, *(src + i));
+      }
+  }
+
+  /*
+   * Memory Op Wrapper Functions
+   */
+
+  void EthereumInterface::loadUint256(uint32_t srcOffset, struct evm_uint256be *dst)
```
axic: These should use constant references `const&` instead of pointers.

src/eei.cpp
```diff
@@ -119,12 +118,108 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+  {
+      assert(dst != nullptr);
+      loadMemory(srcOffset, dst->bytes, 32);
+      endianSwap(dst->bytes, 32);
+  }
+
+  void EthereumInterface::storeUint256(uint32_t dstOffset, struct evm_uint256be *src)
```
axic: These should use constant references `const&` instead of pointers.

src/eei.cpp
jakelang: This would break the load operations, since the function writes to the object referred to by the pointer.

src/eei.cpp
jakelang: Constant references break the endianness switching in all of the wrapper functions. If this is removed, endianness switching must be done manually in all of the interface methods.

src/eei.cpp
```diff
@@ -119,12 +118,96 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      heraAssert(!exceedsUint128(src), "Value at src cannot exceed 2^128-1");
+      storeMemory(dstOffset, src.bytes, 16);
+  }
+
+  /*
+   * Utilities
+   */
+
+  /* Checks if host supplied 256 bit value exceeds UINT128_MAX */
+  unsigned int EthereumInterface::exceedsUint128(const evm_uint256be &value)
```
axic: Will make this `static` (or at least `const`) as it is not reading from the class instance.

PR #22, opened 2018-01-10
Add CMake and build Binaryen as ExternalProject

chfast:


axic:
Can you just create band new commits instead keeping the old one? It doesn't add any benefit it since it changed so much.

PR #28, opened 2018-01-11
Final step of C++ cleanup

axic:


src/eei.h
```diff
@@ -50,8 +51,22 @@ class InternalErrorException : std::exception {
...
 
+struct ExecutionResult {
+  uint64_t gasLeft;
+  std::vector<char> returnValue;
+};
+
 struct EthereumInterface : ShellExternalInterface {
-  EthereumInterface(Hera const& _hera, HeraCall & _call) : ShellExternalInterface(), hera(_hera), call(_call) { (void)hera; }
+  EthereumInterface(
+    struct evm_context const& _context,
```
chfast: You don't need `struct` prefix in C++.

src/eei.h
```diff
@@ -50,8 +51,22 @@ class InternalErrorException : std::exception {
...
+    struct evm_message const& _msg,
+    ExecutionResult & _result
+  ):
+    ShellExternalInterface(),
+    context(_context),
+    msg(_msg),
+    result(_result)
+  { (void)context; }
```
chfast: The `(void)` is not needed.

src/hera.cpp
```diff
@@ -109,7 +160,7 @@ static void evm_destroy(struct evm_instance* instance)
   free(instance);
 }
 
-struct evm_instance* hera_create()
+struct evm_instance* hera_create(void)
```
chfast: The `void` is only needed in declarations.

src/eei.h
axic: Needed to silence unused variable warning (soon it will be used).

PR #37, opened 2018-01-12
EEI: getBalance interface method

jakelang:


src/eei.cpp
```diff
@@ -61,6 +61,25 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("getBalance")) {
+      std::cout << "getbalance\n";
+
+      uint32_t addressOffset = arguments[0].geti32();
+      uint32_t resultOffset = arguments[1].geti32();
+
+      evm_address *address = new evm_address;
```
axic: These can be local variables (no need for pointers & new) because `loaduint160` takes a reference.

src/eei.cpp
```diff
@@ -61,6 +61,25 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      evm_uint256be *result = new evm_uint256be;
```
axic: If `result` is passed to C as a reference this doesn't need to be allocated either.

PR #39, opened 2018-01-12
EEI: interface method call, callCode, callDelegate and callStatic

jakelang:


src/eei.cpp
```diff
@@ -75,6 +75,46 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      context->fn_table->call(&call_result, context, &call_message);
+
+      delete input_data;
+      storeMemory(call_result.output_data, resultOffset, resultLength);
+      vm_status = call_result.status_code;
+
+      if (call_result.release != nullptr)
+        call_result.release(&call_result);
+
+      return Literal((int32_t)vm_status);
```
axic: No need to typecast here, it is the same type.

src/eei.cpp
```diff
@@ -75,6 +75,46 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      int32_t vm_status = 0;
+      uint8_t *input_data = new uint8_t[dataLength];
+
+      call_message.address = loadUint160(addressOffset);
+      call_message.sender = msg.address;
+      call_message.value = loadUint128(valueOffset);
+      loadMemory(dataOffset, input_data, dataLength);
+      call_message.input = input_data;
+      call_message.input_size = dataLength;
+      std::memset(&call_message.code_hash, 0, sizeof(evm_uint256be));
```
axic: `std` is imported

src/eei.cpp
```diff
@@ -90,6 +90,74 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      call_message.depth = msg.depth + 1;
+
+      context->fn_table->call(&call_result, context, &call_message);
+
+      storeMemory(call_result.output_data, resultOffset, resultLength);
+
+      if (call_result.release != nullptr)
+        call_result.release(&call_result);
+
+      return Literal((!call_result.status_code) ? 0 : 1);
```
axic: Should be `(status_code == EVM_SUCCESS) ? 1 : 0`.

src/eei.cpp
```diff
@@ -90,6 +90,74 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("call") ||
+        import->base == Name("callCode") ||
+	import->base == Name("callDelegate") ||
+	import->base == Name("callStatic")) {
```
axic: Trying to follow the coding style of Solidity, so this should be:
```
if (
  import->base .. ||
  import->base ...
) {
```

src/eei.cpp
```diff
@@ -90,6 +90,76 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      loadMemory(dataOffset, input_data, dataLength);
+      call_message.input = input_data.data();
+      call_message.input_size = dataLength;
+      call_message.code_hash = { };
+      call_message.gas = gas;
+      call_message.depth = msg.depth + 1;
+
+      context->fn_table->call(&call_result, context, &call_message);
+
+      storeMemory(call_result.output_data, resultOffset, resultLength);
```
axic: This should use `storeMemory(vector<uint8_t>(call_result.output_data, call_result.output_data + output_size)` to avoid byteswapping.

We'll optimize this later.

src/eei.cpp
```diff
@@ -90,6 +90,76 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+
```
axic: Need to also clear the wasm memory here for `abs(resultLength, output_size)`, e.g. the case where the wasm contract asks for 32 bytes but the call only returns 30 bytes, the top 2 have to be zeroed out according to the semantics.

src/eei.cpp
axic: Also put this into a `if (call_result.output_data)` conditional. See https://github.com/ethereum/evmjit/pull/152/files.

jakelang: we can re-introduce the endianSwap helper function and do manual endianness swapping, since the byte-swapping integrated into memory ops has been more annoying than helpful

src/eei.cpp
```diff
@@ -90,6 +90,76 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      if (call_result.release != nullptr)
```
chfast: This is not bad but I decided to follow [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) in open source projects, mostly to cut unproductive discussions and not to craft "coding style" all over again.

I have no read it all, but for this case I've fount a recommendation to not use `!= nullptr`:
https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es87-dont-add-redundant--or--to-conditions

src/eei.cpp
```diff
@@ -90,6 +90,80 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      call_message.code_hash = { };
+      call_message.gas = gas;
+      call_message.depth = msg.depth + 1;
+
+      context->fn_table->call(&call_result, context, &call_message);
+
+      if (call_result.output_data) {
+      	vector<uint8_t> result(resultLength);
+	for (size_t i = 0; i < resultLength; ++i) {
+	    result[i] = (i < call_result.output_size) ? call_result.output_data[i] : 0;
```
axic: Apparently we could do:
```
vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
// NOTE: this will grow the buffer (and zero pad) or shrink (and truncate)
result.resize(resultLength);
```

src/eei.cpp
```diff
@@ -90,6 +90,78 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      	valueOffset = arguments[2].geti32();
+	dataOffset = arguments[3].geti32();
+	dataLength = arguments[4].geti32();
+	resultOffset = arguments[5].geti32();
+	resultLength = arguments[6].geti32();
+
+	call_message.sender = msg.address;
+	call_message.value = loadUint128(valueOffset);
+	call_message.kind = (import->base == Name("callCode")) ? EVM_CALLCODE : EVM_CALL;
+	call_message.flags = 0;
```
axic: Apparently we need to copy the static flag. I suggest we do that outside this `if`:
```
heraAssert((msg.flags & ~EVM_STATIC) == 0, "Unknown flags not supported.");
call_message.flags = msg.flags;
```

And then do `call_message.flags |= EVM_STATIC` within the static call only.

chfast: Check out this change in EVMJIT: https://github.com/ethereum/evmjit/commit/2f75569213e5710d096aec1976c1add2d0be3438#diff-5719df190800af8d47396050a508080e.

src/eei.cpp
```diff
@@ -90,6 +90,78 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+	dataLength = arguments[3].geti32();
+	resultOffset = arguments[4].geti32();
+	resultLength = arguments[5].geti32();
+
+	if (import->base == Name("callDelegate")) {
+	  call_message.sender = msg.sender;
+	  call_message.value = msg.value;
+	  call_message.kind = EVM_DELEGATECALL;
+	  call_message.flags = 0;
+	} else {
```
chfast: I would got with `else if (import->base == Name("callStatic")` to be able to find it's implementation quickly.

src/eei.cpp
```diff
@@ -90,6 +90,80 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      if (call_result.output_data) {
+      	 vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
+	 result.resize(resultLength);
+	 storeMemory(result, 0, resultOffset, resultLength);
+      }
+
+      if (call_result.release)
+        call_result.release(&call_result);
+
+      return Literal((call_result.status_code == EVM_SUCCESS) ? 1 : 0);
```
jakelang: @axic would this not be the other way around? The spec states that this should return 1 when the VM traps, and 0 otherwise.

axic: Oh probably we fixed it in eWASM. EVM has a bool. In that case eWASM should return states properly: 0 (success), 1 (fail), 2 (revert)

axic:
@chfast please re-review

src/eei.cpp
```diff
@@ -157,6 +157,89 @@ namespace HeraVM {
...
+      call_message.input = input_data.data();
+      call_message.input_size = dataLength;
+
+      evm_result call_result;
+      context->fn_table->call(&call_result, context, &call_message);
+
+      if (call_result.output_data) {
+        vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
+        result.resize(resultLength);
+        storeMemory(result, 0, resultOffset, resultLength);
```
chfast: This will not work when `resultLength` > `call_result.output_size`. The missing part will not be filed with zeros.

PR #40, opened 2018-01-13
Run evm2wasm on EVM input

axic:


axic:
@chfast can you add a cmake rule to compile `evm2wasm` and include its root for the header? Note: `evm2wasm` doesn't have a header, yet - I'll fix that later.

chfast:
It needs most of the changes to be done in evm2wasm. Let me know when they are ready to check the CMake there.

axic:
@chfast evm2wasm stuff is merged

src/hera.cpp
```diff
@@ -52,6 +54,12 @@ void execute(
 ) {
   cout << "Executing...\n";
 
+  // Translate EVM bytcode to WASM
+  if (code[0] != 0 || code[1] != 'a' || code[2] != 's' || code[3] != 'm') {
+    std::string str_code = evm2wasm(std::string{code.data(), code.size()});
```
axic: Has `using std`.

axic:
Can you add `-DEVM2WASM=1` option to cmake?

src/hera.cpp
axic: Would name it actually `translated`.

src/hera.cpp
```diff
@@ -52,6 +54,12 @@ void execute(
 ) {
   cout << "Executing...\n";
 
+  // Translate EVM bytcode to WASM
+  if (code[0] != 0 || code[1] != 'a' || code[2] != 's' || code[3] != 'm') {
```
axic: Need to check for `code.size() >= 4`.

src/hera.cpp
```diff
@@ -52,6 +56,14 @@ void execute(
 ) {
   cout << "Executing...\n";
 
+#if HERA_EVM2WASM
+  // Translate EVM bytecode to WASM
+  if ((code.size() >= 4) && (code[0] != 0 || code[1] != 'a' || code[2] != 's' || code[3] != 'm')) {
```
axic: Ah this should be if `code.size() < 4 || ...`

axic:
@chfast do you want to review this? It is not important, won't be merged soon.

PR #44, opened 2018-01-13
EEI: interface method "codeCopy"

jakelang:
uses context fix from eei-getbalance

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -115,6 +115,18 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("codeCopy")) {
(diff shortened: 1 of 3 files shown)


src/eei.cpp
```diff
@@ -81,6 +95,21 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
     }
 
+    if (import->base == Name("codeCopy")) {
+      cout << "codecopy\n";
+
+      uint32_t resultOffset = arguments[0].geti32();
+      uint32_t codeOffset = arguments[1].geti32();
+      uint32_t length = arguments[2].geti32();
+
+      vector<uint8_t> code(length);
```
axic: This is actually copying the currently executing code. Changing that because it requires a couple of changes.

PR #45, opened 2018-01-13
EEI: interface method "getCodeSize"

jakelang:
uses context fix from eei-getbalance

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -127,6 +127,12 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("getCodeSize")) {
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -81,6 +95,14 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("getCodeSize")) {
+      cout << "codesize\n";
+
+      size_t code_size = context->fn_table->get_code(NULL, context, &msg.address);
```
axic: This code is nice, but unfortunately it will be used in `getExternalCodeSize`. For `getCodeSize` we have it locally as `code.size()`.

PR #48, opened 2018-01-13
EEI: interface method "getGasLeft"

jakelang:


axic:
This was actually implemented by #33.

PR #59, opened 2018-01-14
EEI: implement externalCodeCopy

axic:


src/eei.cpp
```diff
@@ -133,6 +133,25 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      uint32_t codeOffset = arguments[2].geti32();
+      uint32_t length = arguments[3].geti32();
+
+      evm_address address = loadUint160(addressOffset);
+      const uint8_t *code;
+      size_t code_size = context->fn_table->get_code(&code, context, &address);
+
+      // NOTE: code will be freed by the call (client)
+
+      storeMemory(vector<uint8_t>(code, code + code_size), codeOffset, resultOffset, length);
```
axic: We create a vector here for simplicity only (`storeMemory` with checks operates on it). Will optimise this later.

src/eei.cpp
```diff
@@ -133,6 +133,25 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      cout << "externalcodecopy" << endl;
+
+      uint32_t addressOffset = arguments[0].geti32();
+      uint32_t resultOffset = arguments[1].geti32();
+      uint32_t codeOffset = arguments[2].geti32();
+      uint32_t length = arguments[3].geti32();
+
+      evm_address address = loadUint160(addressOffset);
+      const uint8_t *code;
+      size_t code_size = context->fn_table->get_code(&code, context, &address);
```
axic: @chfast `get_code` returns a pointer to the pointer, not clear who needs to free that

src/eei.cpp
```diff
@@ -133,6 +133,25 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+
+      // NOTE: code will be freed by the call (client)
```
axic: `callee`

PR #61, opened 2018-01-14
Support metering in create interface and create tx

axic:


axic:
@chfast do you want to review how this could be done with way less copying of code?

src/hera.cpp
```diff
@@ -44,6 +44,14 @@ using namespace HeraVM;
 
 namespace {
 
+vector<uint8_t> sentinel(vector<uint8_t> const& input)
```
chfast: For this case you can pass by value, but that will probably change with the real metering.

src/hera.cpp
```diff
@@ -130,15 +138,33 @@ static struct evm_result evm_execute(
...
     execute(context, _code, *msg, result);
 
     // copy call result
     if (result.returnValue.size() > 0) {
-      ret.output_size = result.returnValue.size();
+      vector<uint8_t> returnValue;
+
+      if (msg->kind == EVM_CREATE) {
+        // Meter the deployed code
+        vector<uint8_t> metered = sentinel(reinterpret_cast<vector<uint8_t> const&>(result.returnValue));
```
chfast: You can move the `result.returnValue` but `reinterpret_cast` must be eliminated first.

src/hera.cpp
```diff
@@ -130,15 +138,33 @@ static struct evm_result evm_execute(
...
+        heraAssert(metered.size() > 5, "Invalid contract or metering failed.");
+        returnValue = metered;
+      } else {
+        returnValue = reinterpret_cast<vector<uint8_t> &>(result.returnValue);
```
chfast: You can also move here.

src/hera.cpp
```diff
@@ -130,15 +171,32 @@ static struct evm_result evm_execute(
...
+    }
+
     execute(context, _code, *msg, result);
 
     // copy call result
     if (result.returnValue.size() > 0) {
-      ret.output_size = result.returnValue.size();
+      vector<uint8_t> returnValue;
+
+      if (msg->kind == EVM_CREATE && !result.isRevert) {
```
chfast: Does the `.isRevert`  also mean a "failure"?

src/hera.cpp
```diff
@@ -130,15 +171,32 @@ static struct evm_result evm_execute(
...
+      } else {
+        returnValue = move(result.returnValue);
+      }
+
+      ret.output_size = returnValue.size();
       ret.output_data = (const uint8_t *)malloc(ret.output_size);
       heraAssert(ret.output_data != NULL, "Memory allocation failure.");
       ret.release = evm_destroy_result;
-      copy(result.returnValue.begin(), result.returnValue.end(), (char *)ret.output_data);
+      copy(returnValue.begin(), returnValue.end(), (char *)ret.output_data);
```
chfast: `(char*)` -> `reinterpret_cast`.

src/hera.cpp
```diff
@@ -44,6 +44,47 @@ using namespace HeraVM;
...
 
+vector<uint8_t> sentinel(struct evm_context* context, vector<uint8_t> const& input)
+{
+#if HERA_DEBUGGING
+  cout << "Metering (input " << input.size() << " bytes)..." << endl;
+#endif
+
+#if HERA_METERING_CONTRACT
+  evm_message metering_message = {
+    .address = { .bytes = { 10 } }, // precompile address 0x00...0a
```
chfast: The address is an array of bytes (aka "big endian"), so it must be `{0, 0, 0, ... , 10}`.

src/hera.cpp
axic: Yes. Real failures are caught by the exception block, here only two cases can happen success or revert.

lrettig:
My tests (https://github.com/ewasm/tests/pull/2) pass with this change

PR #62, opened 2018-01-14
Remove obsolete features from ShellInterface

axic:


axic:
Do not merge this yet.

axic:
@chfast ready for review?

src/eei.cpp
```diff
@@ -56,6 +56,12 @@ string toHex(evm_uint256be const& value) {
 
 }
 
+  void EthereumInterface::importGlobals(std::map<Name, Literal>& globals, Module& wasm) {
+    (void)globals;
+    (void)wasm;
+    HERA_DEBUG << "importGlobal " << "\n";
```
chfast: Why 2 literals?

axic: Because there was `<< globals <<` in between :)

If you know an easy way to dump a map that would be nice.

chfast: Only `for` loop.

axic: So should this be updated?

axic:
@chfast can this be merged now?

PR #65, opened 2018-01-15
EEI: charge for memory gas cost in useGas

axic:


axic:
Depends on decision in https://github.com/ewasm/design/pull/75

src/eei.cpp
```diff
@@ -199,6 +199,9 @@ string toHex(evmc_uint256be const& value) {
 
       takeGas(gas);
 
+      heraAssert((ffsl(gas) + ffsl(memory.size()) <= 64), "Memory gas calculation overflow."); //may need to find alternative to ffsl for cross-libc portability
+      takeGas(gas * memory.size() / GasSchedule::memoryPageSize * GasSchedule::memoryCostPerPage + 1); //round gas cost up
```
jakelang: Unsure whether this should be rounded up or not.

axic: Not sure. What are the pros/cons for each?

jakelang:
rebased

PR #67, opened 2018-01-15
EEI: interface method log

jakelang:


src/eei.cpp
```diff
@@ -224,6 +224,29 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      uint32_t numberOfTopics = arguments[2].geti32();
+
+      uint8_t data[length];
+      evm_uint256be topics[numberOfTopics];
+      uint32_t topicOffset;
+
+      loadMemory(dataOffset, data, length);
+
+      for (size_t i = 0; i < numberOfTopics; ++i) {
+        topicOffset = arguments[i + 3].geti32();
```
axic: Can you do this as `3 + i` ?

src/eei.cpp
```diff
@@ -224,6 +224,29 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
     }
 
+    if (import->base == Name("log")) {
+      cout << "log" << endl;
+
+      uint32_t dataOffset = arguments[0].geti32();
+      uint32_t length = arguments[1].geti32();
+      uint32_t numberOfTopics = arguments[2].geti32();
+
+      uint8_t data[length];
```
axic: I think we have to avoid allocating large data on the stack. Can you go for the moment with the messy `vector` allocation (also avoids the byteswapping)?

src/eei.cpp
axic: Should also move the variable declaration here for `topicOffset`.

src/eei.cpp
```diff
@@ -224,6 +224,29 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      uint32_t dataOffset = arguments[0].geti32();
+      uint32_t length = arguments[1].geti32();
+      uint32_t numberOfTopics = arguments[2].geti32();
+
+      vector<uint8_t> data(length);
+      evm_uint256be topics[numberOfTopics];
+      uint32_t topicOffset;
+
+      for (size_t i = 0; i < numberOfTopics; ++i) {
+        topicOffset = arguments[3 + i].geti32();
```
axic: Can you move the `topicOffset` declaration here?

src/eei.cpp
```diff
@@ -224,6 +224,29 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+        topics[i] = loadUint256(topicOffset);
+      }
+
+      loadMemory(dataOffset, data, length);
```
axic: Can you move the vector declaration to here?

PR #69, opened 2018-01-15
 EEI: add better debugging option

axic:


src/eei.cpp
```diff
@@ -28,17 +28,43 @@
...
+  stringstream s;
+  s << hex;
+  for (unsigned i = 0; i < 32; i++)
+    s << value.bytes[i];
+  return s.str();
+}
+
+}
+
+#if HERA_DEBUGGING
```
lrettig: Isn't this conditional backwards? You want to print if HERA_DEBUGGING is ON right?

PR #71, opened 2018-01-15
EEI: add gas costs for storageStore

axic:


src/eei.cpp
```diff
@@ -284,6 +284,16 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       evm_uint256be path = loadUint256(pathOffset);
       evm_uint256be value = loadUint256(valueOffset);
 
+      evm_uint256be current;
+      context->fn_table->get_storage(&current, context, &msg.address, &path);
+
+      // We do not need to take care about the delete case, the client does it.
```
chfast: More precisely, you don't have to care about refunding.

src/eei.cpp
```diff
@@ -440,11 +450,19 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
    * Utilities
    */
 
-  /* Checks if host supplied 256 bit value exceeds UINT128_MAX */
   unsigned int EthereumInterface::exceedsUint128(evm_uint256be const& value)
   {
-    for (unsigned i = 0; i < 16; ++i) {
-      if (value.bytes[i])
+    for (unsigned i = 0; i < 16; i++) {
```
chfast: You can implement it as:
```
return std::any_of(&value.bytes[0], &value.bytes[16], [](uint8_t b) { return b != 0; });
```
But not forcing it...

src/eei.h
```diff
@@ -92,12 +92,21 @@ struct EthereumInterface : ShellExternalInterface {
   evm_uint256be loadUint128(uint32_t srcOffset);
   void storeUint128(evm_uint256be const& src, uint32_t dstOffset);
 
-  unsigned int exceedsUint128(evm_uint256be const& value);
+  /* Checks if host supplied 256 bit value exceeds UINT128_MAX */
+  static unsigned int exceedsUint128(evm_uint256be const& value);
```
chfast: Why it returns `int`?

axic: Yeah should be bool

src/eei.cpp
axic: Can you open a new PR with that change?

PR #72, opened 2018-01-15
Implement debug namespace

### LINKED ISSUE #70: Implement the debug namespace
Should add the same methods as ewasm-kernel: https://github.com/ewasm/ewasm-kernel/blob/master/debugInterface.js

- `print32(value: i32)`
- `print64(value: i64)`
- `printMem(offset: i32, len: i32)` - print memory segment as printable characters
- `printMemHex(offset: i32, len: i32)` - print memory segment as a hex

Should only be enabled in debug mode.

lrettig:
Fixes #70.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,5 +7,9 @@ add_library(hera
     ${evm_include_dir}/evm.h
 )
 
+if(HERA_DEBUGGING)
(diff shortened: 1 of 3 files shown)


lrettig:
Source: https://github.com/ewasm/hera/issues/70

lrettig:
Closing in favor of #75

axic:
There was no need to close it, could have force pushed 😉

PR #76, opened 2018-01-15
EEI: interface method create

jakelang:


src/eei.cpp
```diff
@@ -191,6 +191,35 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      cout << "create" << endl;
+
+      uint32_t valueOffset = arguments[0].geti32();
+      uint32_t dataOffset = arguments[1].geti32();
+      uint32_t length = arguments[2].geti32();
+      uint32_t resultOffset = arguments[3].geti32();
+
+      evm_message create_message;
+      evm_result create_result;
+      vector<uint8_t> contract_code(length);
```
axic: Please move this before the `loadMemory`.

src/eei.cpp
```diff
@@ -191,6 +191,35 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+    if (import->base == Name("create")) {
+      cout << "create" << endl;
+
+      uint32_t valueOffset = arguments[0].geti32();
+      uint32_t dataOffset = arguments[1].geti32();
+      uint32_t length = arguments[2].geti32();
+      uint32_t resultOffset = arguments[3].geti32();
+
+      evm_message create_message;
+      evm_result create_result;
```
axic: Please move this before the `call` fn.

src/eei.cpp
```diff
@@ -191,6 +191,35 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
       return Literal();
     }
 
+    if (import->base == Name("create")) {
+      cout << "create" << endl;
```
axic: Please use the new debugging helper.

src/eei.cpp
```diff
@@ -191,6 +191,35 @@ Literal EthereumInterface::callImport(Import *import, LiteralList& arguments) {
...
+      create_message.sender = msg.address;
+      create_message.value = loadUint128(valueOffset);
+      loadMemory(dataOffset, contract_code, length);
+      create_message.input = contract_code.data();
+      create_message.input_size = length;
+      create_message.code_hash = { };
+      create_message.depth = msg.depth + 1;
+      create_message.kind = EVM_CREATE;
+      create_message.flags = 0;
+
```
axic: `gas` must be calculated properly here according to the weird rules we have ...

`gasLeft - (gasLeft / 64)`

src/eei.cpp
```diff
@@ -479,6 +479,41 @@ namespace HeraVM {
...
+      evm_result create_result = { };
+      context->fn_table->call(&create_result, context, &create_message);
+      storeUint160(create_result.create_address, resultOffset);
+
+      if (create_result.release)
+        create_result.release(&create_result);
+
+      return Literal((create_result.status_code == EVM_SUCCESS) ? true : false);
```
axic: I think this should be 1 and 0 - perhaps binaryen has an overload for boolean but better to avoid it.

Oh and also please typecast the entire thing to `uint32_t` (we forgot to do that for calls...).

Just to be explicit.

PR #78, opened 2018-01-16
Fix loadUint128()

chfast:
The half of returned value was not initialized.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -475,7 +475,7 @@ namespace HeraVM {
 
   evm_uint256be EthereumInterface::loadUint256(uint32_t srcOffset)
   {
-    evm_uint256be dst;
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -475,31 +475,31 @@ namespace HeraVM {
...
   void EthereumInterface::storeUint160(evm_address const& src, uint32_t dstOffset)
   {
-    storeMemory(src.bytes, dstOffset, 20);
+    storeMemory(src.bytes, dstOffset, sizeof(src));
   }
 
   evm_uint256be EthereumInterface::loadUint128(uint32_t srcOffset)
   {
-    evm_uint256be dst;
+    evm_uint256be dst = {};
```
chfast: The bug is only here.

PR #80, opened 2018-01-16
 EEI: implement getReturnDataSize and returnDataCopy

axic:


src/eei.cpp
```diff
@@ -468,9 +477,9 @@ namespace HeraVM {
       context->fn_table->call(&call_result, context, &call_message);
 
       if (call_result.output_data) {
-        vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
-        result.resize(resultLength);
-        storeMemory(result, 0, resultOffset, resultLength);
+        lastReturnData.assign(call_result.output_data, call_result.output_data + call_result.output_size);
```
jakelang: Is call not supposed to store output data at resultOffset anyway?

axic: There is no point doing that with returndata anymore. See https://github.com/ewasm/design/issues/70

axic: But you are correct, perhaps I should separate these changes into two: supporting returndata and removing that option.

axic: Done, that will be a different PR.

src/eei.cpp
```diff
@@ -401,6 +401,23 @@ namespace HeraVM {
       return Literal();
     }
 
+    if (import->base == Name("returnDataSize")) {
+      HERA_DEBUG << "returnDataSize\n";
+      return Literal((uint32_t)lastReturnData.size());
```
chfast: Use `static_cast`.

src/eei.cpp
```diff
@@ -475,6 +492,10 @@ namespace HeraVM {
         vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
         result.resize(resultLength);
         storeMemory(result, 0, resultOffset, resultLength);
+
+        lastReturnData.assign(call_result.output_data, call_result.output_data + call_result.output_size);
+      } else {
+        lastReturnData.resize(0);
```
chfast: 1. `.resize(0)` === `.clear()`.
2. You have to clear it unconditionally.

src/eei.h
```diff
@@ -104,6 +104,7 @@ struct EthereumInterface : ShellExternalInterface {
   struct evm_context* context = nullptr;
   std::vector<uint8_t> const& code;
   struct evm_message const& msg;
+  std::vector<uint8_t> lastReturnData = std::vector<uint8_t>();
```
chfast: You need a default initializer only for "primitive" types.

src/eei.cpp
```diff
@@ -475,6 +492,10 @@ namespace HeraVM {
         vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
         result.resize(resultLength);
         storeMemory(result, 0, resultOffset, resultLength);
+
+        lastReturnData.assign(call_result.output_data, call_result.output_data + call_result.output_size);
```
chfast: The same goes for CREATE.

src/eei.cpp
```diff
@@ -401,6 +401,23 @@ namespace HeraVM {
       return Literal();
     }
 
+    if (import->base == Name("returnDataSize")) {
```
axic: Should be `getReturnDataSize`.

PR #81, opened 2018-01-16
EEI: specify explicit return types

axic:


src/eei.cpp
```diff
@@ -275,7 +275,7 @@ namespace HeraVM {
       evm_tx_context tx_context;
       context->fn_table->get_tx_context(&tx_context, context);
 
-      return Literal((int64_t)tx_context.block_gas_limit);
+      return Literal((uint64_t)tx_context.block_gas_limit);
```
chfast: Use `static_cast`.

src/eei.cpp
```diff
@@ -317,7 +317,7 @@ namespace HeraVM {
       evm_tx_context tx_context;
       context->fn_table->get_tx_context(&tx_context, context);
 
-      return Literal(tx_context.block_number);
+      return Literal((uint64_t)tx_context.block_number);
```
chfast: Use `static_cast`.

src/eei.cpp
```diff
@@ -326,7 +326,7 @@ namespace HeraVM {
       evm_tx_context tx_context;
       context->fn_table->get_tx_context(&tx_context, context);
 
-      return Literal(tx_context.block_timestamp);
+      return Literal((uint64_t)tx_context.block_timestamp);
```
chfast: Use `static_cast`.

src/eei.cpp
```diff
@@ -476,7 +476,7 @@ namespace HeraVM {
       if (call_result.release)
         call_result.release(&call_result);
 
-      return Literal((call_result.status_code == EVM_SUCCESS) ? 1 : 0);
+      return Literal((call_result.status_code == EVM_SUCCESS) ? (uint32_t)1 : (uint32_t)0);
```
chfast: Is this cast needed? Compilers usually handle literals nicely.
If needed, write it as `uint32_t(1) : uint32_t(0)`.

axic: Needed because it has a lot of overloads and this is to ensure it is always treated as 32 bit.

src/eei.cpp
axic: Actually these can stay `int64_t`, but is there a way to just throw an error if they are not `int64_t` without typecasting?

PR #84, opened 2018-01-16
EEI: change call to not return result data

axic:
See https://github.com/ewasm/design/issues/70

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -468,8 +468,6 @@ namespace HeraVM {
       uint32_t valueOffset;
       uint32_t dataOffset;
       uint32_t dataLength;
-      uint32_t resultOffset;
-      uint32_t resultLength;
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -503,10 +495,6 @@ namespace HeraVM {
       context->fn_table->call(&call_result, context, &call_message);
 
       if (call_result.output_data) {
-        vector<uint8_t> result(call_result.output_data, call_result.output_data + call_result.output_size);
-        result.resize(resultLength);
-        storeMemory(result, 0, resultOffset, resultLength);
-
         lastReturnData.assign(call_result.output_data, call_result.output_data + call_result.output_size);
```
chfast: You can do it unconditionally now. It will handle case of (null, 0) as well.

axic: I think I'll leave it for the moment. Do you want to create a new PR which changes to in all appropriate places?

PR #85, opened 2018-01-16
Adds printStorage and printStorageHex

lrettig:


src/eei.cpp
```diff
@@ -92,6 +92,40 @@ namespace HeraVM {
...
+      uint32_t pathOffset = arguments[0].geti32();
+      evm_uint256be path = loadUint256(pathOffset);
+
+      bool useHex = import->base == Name("printStorageHex");
+
+      HERA_DEBUG << "DEBUG printStorage" << (useHex ? "Hex" : "") << "(0x" << hex;
+
+      // Print out the path
+      for (uint8_t b : path.bytes)
+        cout << static_cast<int>(b);
```
jakelang: is the cast needed here?

src/eei.cpp
```diff
@@ -92,6 +92,40 @@ namespace HeraVM {
...
+
+      HERA_DEBUG << "): " << dec;
+
+      evm_uint256be result;
+      context->fn_table->get_storage(&result, context, &msg.address, &path);
+
+      if (useHex)
```
jakelang: you could factor this down into something like
```
cout << (useHex) ? hex : dec;
for (uint8_t b : result.bytes)
    cout << b << " ";
cout << dec << endl;

lrettig:
Thanks for updating the README here, sorry I forgot to do that!

PR #86, opened 2018-01-16
Charge gas for EEI methods

jakelang:


src/eei.h
```diff
@@ -108,8 +108,18 @@ struct EthereumInterface : ShellExternalInterface {
 };
 
 struct GasSchedule {
+  static constexpr unsigned sLoad = 50;
```
axic: Should be called `storageLoad`

src/eei.h
```diff
@@ -108,8 +108,18 @@ struct EthereumInterface : ShellExternalInterface {
...
   static constexpr unsigned storageStoreCreate = 20000;
   static constexpr unsigned storageStoreChange = 5000;
+  static constexpr unsigned log = 375;
+  static constexpr unsigned logData = 8;
+  static constexpr unsigned logTopic = 375;
+  static constexpr unsigned create = 32000;
+  static constexpr unsigned call = 40;
+  static constexpr unsigned memory = 3;
+  static constexpr unsigned copy = 3;
```
axic: What are these two used for?

jakelang: They are used in codecopy and calldatacopy, according to outdated yellowpaper semantics. Will fix this

axic: Can you name perhaps have `codeMemory`, `codeCopy` and `callDataMemory`, `callDataCopy`?

jakelang: yeah, I was using the fee schedule names from EVMJIT and the yellowpaper but this makes more sense i guess.

src/eei.cpp
```diff
@@ -209,6 +221,8 @@ namespace HeraVM {
       vector<uint8_t> input(msg.input, msg.input + msg.input_size);
       storeMemory(input, dataOffset, resultOffset, length);
 
+      takeGas(GasSchedule::stepGas2 + GasSchedule::copy * (length / 32));
```
chfast: @axic What do you think about charging for memory per byte, instead of per "word"?
@jakelang I believe `length / 32` should be rounded up, not down.

src/eei.cpp
```diff
@@ -241,12 +259,16 @@ namespace HeraVM {
 
       storeMemory(code, codeOffset, resultOffset, length);
 
+      takeGas(GasSchedule::stepGas2 + GasSchedule::copy * (length / 32));
```
chfast: Same here.

src/eei.cpp
```diff
@@ -347,6 +382,8 @@ namespace HeraVM {
 
       context->fn_table->log(context, &msg.address, data.data(), length, topics, numberOfTopics);
 
+      takeGas(GasSchedule::log + (length * GasSchedule::logData) + (GasSchedule::logTopic * numberOfTopics));
```
chfast: How do you prevent overflows?

src/eei.cpp
```diff
@@ -550,6 +602,14 @@ namespace HeraVM {
       if (call_result.release)
         call_result.release(&call_result);
 
+      if (!isZeroUint256(call_message.value))
+        takeGas(GasSchedule::valuetransfer);
+
+      if (import->base == Name("call") && !context->fn_table->account_exists(context, &call_message.address))
+        takeGas(GasSchedule::callnewacct);
+
+      takeGas(GasSchedule::call);
```
chfast: 1. This should go first, before other checks.
2. Also, the gas should be taken _before_ the actual call.
3. You also have to take gas of the gas argument, and put back the amount left after the call.

src/eei.cpp
```diff
@@ -601,6 +661,9 @@ namespace HeraVM {
       if (create_result.release)
         create_result.release(&create_result);
 
+      takeGas(GasSchedule::create);
```
chfast: Must go before the CREATE.

src/eei.h
```diff
@@ -116,8 +116,29 @@ struct EthereumInterface : ShellExternalInterface {
...
+  static constexpr unsigned blockhash = 800;
+  static constexpr unsigned balance = 400;
+  static constexpr unsigned stepGas0 = 0;
+  static constexpr unsigned stepGas1 = 2;
+  static constexpr unsigned stepGas2 = 3;
+  static constexpr unsigned stepGas3 = 5;
+  static constexpr unsigned stepGas4 = 8;
+  static constexpr unsigned stepGas5 = 10;
+  static constexpr unsigned stepGas6 = 20;
+  static constexpr unsigned stepGas7 = 0;
```
chfast: Why 0?

axic:
@chfast do you want to review the costs and work on this to finish it?

src/eei.h
jakelang: This was here:
https://github.com/ethereum/evmjit/blob/develop/libevmjit/JIT.h#L71

It is unneeded, so I am getting rid of it

src/eei.cpp
jakelang: I don't see a solution to this that isn't messy.
Maybe a helper function would be better for this.

jakelang:
@chfast I have made appropriate changes. I am also considering writing helpers for arithmetic overflow checking in gas charges.

src/eei.cpp
```diff
@@ -403,6 +438,7 @@ namespace HeraVM {
         GasSchedule::storageStoreChange
       );
 
+      context->fn_table->get_storage(&current, context, &msg.address, &path);
```
axic: This must happen before `takeGas` otherwise it will charge invalidly.

src/eei.cpp
```diff
@@ -550,6 +597,9 @@ namespace HeraVM {
       if (call_result.release)
         call_result.release(&call_result);
 
+      /* Return unspent gas */
+      result.gasLeft += call_result.gas_left;
```
axic: For this to work, the entire passed gas must be used via `takeGas` before doing an EVM-C call.

jakelang: This is done by just performing:
```takeGas(gas);```
before the call execution, correct?

Comment (axic) [reply src/eei.cpp]: Yep

**Comment by axic** [comment]
Please rebase, it is in conflict.

File: src/eei.cpp [hunk 11]
```diff
@@ -470,20 +516,24 @@ namespace HeraVM {
       uint32_t valueOffset;
       uint32_t dataOffset;
       uint32_t dataLength;
+      uint32_t resultOffset;
+      uint32_t resultLength;
```
axic: These shouldn't need to be added back.

src/eei.cpp
```diff
@@ -470,20 +516,24 @@ namespace HeraVM {
...
 
       heraAssert((msg.flags & ~EVM_STATIC) == 0, "Unknown flags not supported.");
 
       evm_message call_message;
       call_message.address = loadUint160(addressOffset);
       call_message.flags = msg.flags;
       call_message.code_hash = {};
-      call_message.gas = gas - (gas / 64);
+      call_message.gas = gas;
```
axic: Please don't remove this.

src/eei.cpp
```diff
@@ -578,11 +646,16 @@ namespace HeraVM {
...
+      takeGas(create_message.gas);
+      takeGas(GasSchedule::create);
+      takeGas(GasSchedule::callnewacct);
       context->fn_table->call(&create_result, context, &create_message);
-      if (create_result.status_code == EVM_SUCCESS) {
+
+      if (create_result.status_code == EVM_SUCCESS)
         storeUint160(create_result.create_address, resultOffset);
-        lastReturnData.clear();
-      } else if (create_result.output_data) {
```
axic: Please don't change this.

src/eei.cpp
```diff
@@ -603,6 +679,7 @@ namespace HeraVM {
 
       evm_address address = loadUint160(addressOffset);
 
+      takeGas(GasSchedule::selfdestruct + (!context->fn_table->account_exists(context, &address)) ? GasSchedule::callnewacct : 0);
```
axic: I'd split this up into two `takeGas` statements for readability.

src/eei.h
```diff
@@ -116,8 +116,28 @@ struct EthereumInterface : ShellExternalInterface {
...
   static constexpr unsigned storageStoreChange = 5000;
+  static constexpr unsigned log = 375;
+  static constexpr unsigned logData = 8;
+  static constexpr unsigned logTopic = 375;
+  static constexpr unsigned create = 32000;
+  static constexpr unsigned call = 700;
+  static constexpr unsigned copy = 3;
+  static constexpr unsigned blockhash = 800;
+  static constexpr unsigned balance = 400;
+  static constexpr unsigned stepGas0 = 0;
```
axic: Seems unused.

src/eei.h
```diff
@@ -116,8 +116,28 @@ struct EthereumInterface : ShellExternalInterface {
...
 struct GasSchedule {
+  static constexpr unsigned storageLoad = 200;
   static constexpr unsigned storageStoreCreate = 20000;
   static constexpr unsigned storageStoreChange = 5000;
+  static constexpr unsigned log = 375;
+  static constexpr unsigned logData = 8;
+  static constexpr unsigned logTopic = 375;
+  static constexpr unsigned create = 32000;
+  static constexpr unsigned call = 700;
+  static constexpr unsigned copy = 3;
```
axic: Which instructions does this copy refer to?

src/eei.h
```diff
@@ -116,8 +116,28 @@ struct EthereumInterface : ShellExternalInterface {
...
+  static constexpr unsigned blockhash = 800;
+  static constexpr unsigned balance = 400;
+  static constexpr unsigned stepGas0 = 0;
+  static constexpr unsigned stepGas1 = 2;
```
axic: I think I'd prefer having named gas fields for each of the interface method as opposed to YellowPaper-esque `stepGas*`

jakelang: any recommendations for a naming scheme?

src/eei.cpp
jakelang: just fixed this, leftover trash from a rebase

src/eei.h
axic: Probably the name of the interface method?

src/eei.h
jakelang: All the copies except externalCodeCopy, hence its general naming.

src/eei.h
jakelang: Lots of these are used for multiple interface methods.

src/eei.h
jakelang: removed now. I kept it there for the sake of correspondence to EVM gas schedule, but this is now unneeded

src/eei.cpp
jakelang: done

src/eei.h
axic: I'd suggest to with the YP in that case: Appendix G of https://ethereum.github.io/yellowpaper/paper.pdf

src/eei.cpp
```diff
@@ -585,7 +643,12 @@ namespace HeraVM {
       create_message.flags = 0;
 
       evm_result create_result;
+     
+      takeGas(create_message.gas);
+      takeGas(GasSchedule::create);
+      takeGas(GasSchedule::callnewacct);
```
chfast: This is not needed here. However, changing GasSchedule::create to be equal GasSchedule::call + GasSchedule::callnewacct would make sense, @axic ?

src/eei.cpp
```diff
@@ -598,6 +661,9 @@ namespace HeraVM {
       if (create_result.release)
         create_result.release(&create_result);
 
+      /* Return unspent gas */
+      result.gasLeft += create_result.gas_left;
```
chfast: I'm not sure this is correct. This was not subtruck  for CREATE case.

src/eei.cpp
```diff
@@ -516,7 +564,7 @@ namespace HeraVM {
         addressOffset << " " <<
         valueOffset << " " <<
         dataOffset << " " <<
-        dataLength << dec << "\n";
+        dataLength << " " << dec << "\n";
```
axic: I think this is not needed. `dec` as misleading it is (perhaps should used explicit `std::dec` here), is switching to decimal display mode for numbers

src/eei.cpp
```diff
@@ -529,6 +577,13 @@ namespace HeraVM {
       }
 
       evm_result call_result;
+
+      if (import->base == Name("call") && !context->fn_table->account_exists(context, &call_message.destination))
+        takeGas(GasSchedule::callnewacct);
+      if (!isZeroUint256(call_message.value))
+        takeGas(GasSchedule::valuetransfer);
+      takeGas(gas);
```
axic: I think this actually should use `call_message.gas` because the all-but-1/64th rule applies to this too.

src/eei.h
```diff
@@ -116,8 +116,27 @@ struct EthereumInterface : ShellExternalInterface {
...
   static constexpr unsigned storageStoreChange = 5000;
+  static constexpr unsigned log = 375;
+  static constexpr unsigned logData = 8;
+  static constexpr unsigned logTopic = 375;
+  static constexpr unsigned create = 32000;
+  static constexpr unsigned call = 700;
+  static constexpr unsigned copy = 3;
+  static constexpr unsigned blockhash = 800;
+  static constexpr unsigned balance = 400;
+  static constexpr unsigned stepGas1 = 2;
```
axic: Can you please use the names from the Yellowpapyer? `veryLow`, `low`, etc.

src/eei.cpp
```diff
@@ -239,6 +254,9 @@ namespace HeraVM {
 
       HERA_DEBUG << "codeCopy " << hex << resultOffset << " " << codeOffset << " " << length << dec << "\n";
 
+      heraAssert(ffs(GasSchedule::copy) + (ffs(length) - 5) <= 64, "Gas charge overflow");
+      heraAssert(numeric_limits<uint64_t>::max() - GasSchedule::stepGas2 >= GasSchedule::copy * (length / 32), "Gas charge overflow");
+      takeGas((uint64_t)ceil(GasSchedule::stepGas2 + GasSchedule::copy * (length / 32)));
```
axic: This one uses `ceil`, but the one in `externalCodeCopy` doesn't. Any reason for that? Also shouldn't this instead just use `(length + 31) / 32` to avoid the use of `ceil`?

axic: It also may make sense creating a helper along the lines of `takeWordCopyGas(baseCost, costPerWord, length)`

src/eei.cpp
```diff
@@ -206,6 +218,7 @@ namespace HeraVM {
 
       HERA_DEBUG << "callDataCopy " << hex << resultOffset << " " << dataOffset << " " << length << dec << "\n";
 
+      takeGas(GasSchedule::stepGas2 + GasSchedule::copy * (length / 32));
```
axic: The rules applied in `codeCopy` and `externalCodeCopy` are not really followed here?

src/eei.cpp
jakelang: This change is leftover from an old version of Hera. May need to make a fresh PR, so that unintended changes like these do not arise.

src/eei.cpp
jakelang: I will do this in a separate PR if there is a significant need for it.

src/eei.h
jakelang: Some of the gas charges in the are unused in EEI. Should they be retained in the gas schedule for the sake of yellowpaper compliance and potential later use, or trimmed off?

axic: Just trimmed off.

src/eei.cpp
```diff
@@ -262,8 +285,10 @@ namespace HeraVM {
       const uint8_t *code;
       size_t code_size = context->fn_table->get_code(&code, context, &address);
 
+      heraAssert(ffs(GasSchedule::copy) + (ffs(length) - 5) <= 64, "Gas charge overflow");
+      heraAssert(numeric_limits<uint64_t>::max() - GasSchedule::extcode >= GasSchedule::copy * (length / 32), "Gas charge overflow");
```
axic: The assertions don't use the same rounding for length?

axic:
@jakelang I hoped that @chfast could review the values but I'll do it tonight so we can stop dragging this PR forever. Will just fix issues as we go along.

src/eei.cpp
```diff
@@ -206,6 +218,10 @@ namespace HeraVM {
 
       HERA_DEBUG << "callDataCopy " << hex << resultOffset << " " << dataOffset << " " << length << dec << "\n";
 
+      heraAssert(ffs(GasSchedule::copy) + (ffs(length) - 5) <= 64, "Gas charge overflow");
+      heraAssert(numeric_limits<uint64_t>::max() - GasSchedule::verylow >= GasSchedule::copy * ((static_cast<uint64_t>(length) + 31) / 32), "Gas charge overflow");
+      takeGas(GasSchedule::verylow + GasSchedule::copy * ((static_cast<uint64_t>(length) + 31) / 32));
```
axic: I really think we should have a helper for this since there 3 copies now.

jakelang: Could open an issue for this and do it in a separate PR, possibly a refactoring pass.
Might as well get it working now.

PR #87, opened 2018-01-16
add debug message when OOG

jwasinger:


src/eei.cpp
```diff
@@ -536,6 +536,7 @@ namespace HeraVM {
   void EthereumInterface::takeGas(uint64_t gas)
   {
     if (gas > result.gasLeft) {
+      HERA_DEBUG << "Out of gas" << "\n";
```
axic: Perhaps change it to `out of gas :(` or add some crazy emoji in unicode

axic: `out of ⛽️ 😭`

PR #89, opened 2018-01-16
 EEI: avoid returning a pointer for msg.input if msg.input_size is 0

jakelang:
Fix bug where an attempt is made to load input data with zero length


axic:
Why do we actually need to assert in `loadMemory` for `> 0`?

PR #95, opened 2018-01-17
Allow zero-length memory load/store

lrettig:


lrettig:
Fixes an issue with #92

PR #99, opened 2018-01-18
Cleanup returnValue handling execute()

axic:


src/hera.cpp
```diff
@@ -134,11 +134,13 @@ static struct evm_result evm_execute(
 
     // copy call result
     if (result.returnValue.size() > 0) {
+      uint8_t* output_data = (uint8_t*)malloc(result.returnValue.size());
```
chfast: You also use `new uint8_t[result.returnValue.size()]`, and `delete[]` in `evm_destroy_result`.

PR #100, opened 2018-01-18
Use new/delete instead of malloc/free

axic:


src/hera.cpp
```diff
@@ -98,9 +98,8 @@ extern "C" {
 
 static void evm_destroy_result(struct evm_result const* result)
 {
-  if (result->output_data) {
-     free((void *)result->output_data);
-  }
+  if (result->output_data)
```
chfast: You also don't need to check for null before delete.

axic:
@chfast updated

axic:
There is also `calloc`/`free` in `hera_create`/`evm_destroy`. Should that also be replaced?

PR #102, opened 2018-01-18
Split out generic callSystemContract

axic:


src/hera.cpp
```diff
@@ -44,38 +44,57 @@ using namespace HeraVM;
...
+  evm_result result;
+  context->fn_table->call(&result, context, &message);
 
   vector<uint8_t> ret;
-  if (metering_result.status_code == EVM_SUCCESS && metering_result.output_data)
-    ret.assign(metering_result.output_data, metering_result.output_data + metering_result.output_size);
+  if (result.status_code == EVM_SUCCESS && result.output_data)
+    ret.assign(result.output_data, result.output_data + result.output_size);
+
+  gas = result.gas_left - gas;
```
chfast: Not `gas = result.gas_left`?

axic: Well I guess that can move to the caller. In `sentinel` I want to display the actual amount used (and later also charge for that).

PR #103, opened 2018-01-19
Validate ECI (contract interface) properties during execution

axic:


axic:
@lrettig @jwasinger have you tried running tests with this?

src/hera.cpp
```diff
@@ -119,6 +119,10 @@ void execute(
   // Validate
   heraAssert(WasmValidator().validate(module), "Module is not valid.");
 
+  // This should be caught during deployment time by the Sentinel.
+  // TODO: validate for other conditions too?
+  heraAssert(module.getExportOrNull(Name("main")) != nullptr, "Contract entry point (\"main\") missing.");
+
```
axic: This should at least validate that the imports are correct: only known imports with the correct number of arguments (perhaps also the types of arguments).

PR #104, opened 2018-01-21
EEI: create should clear the return data buffer on success

axic:


axic:
@lrettig @jwasinger @hugo-dc having a test case for this would be nice:
- do a call which fills the return buffer
- do a successful create
- ensure returndatasize equals to 0

A few more combinations :)

jwasinger:
> ensure returndatasize equals to 0

@axic I'm assuming that you mean to inspect the `returndatabuffer` to ensure it was zeroed out. Otherwise the proposed test case is just testing the behavior of the environment which is already covered by existing state tests.

so the test case would look like this:
  ```
     do CALL which fills returndata buffer....
     let l = length({return data size from call which fills return buffer})
     do CREATE which zeroes returndata buffer....
     assert(mem[returnBuffer:returnBuffer+l] == '0x00000....')

axic:
Actually the only thing which matters is that `returndatasize` returns 0, because `returndatacopy` will throw if someone tries to read more than what's inside (= `returndatasize`).

Note: I am talking here about `getReturnDataSize` and `returnDataCopy`, in all these tests `resultOffset` and `resultLength` should both be zero as they are obsolete and to be removed soon. See https://github.com/ewasm/design/issues/70

jwasinger:
I really would like to have a strong argument for why we can't run state tests transcompiled to eWASM to address cases like this...

I'm not opposed to writing test cases but I feel like we are re-inventing the wheel here

PR #111, opened 2018-01-23
[WIP] Blockhash to signal success or failure 

jwasinger:
Based on proposed spec from https://github.com/ewasm/design/issues/76

*WIP* - Waiting on confirmation of spec changes before more progress will be made


jakelang:
Is this still needed?
@jwasinger

jwasinger:
If we still decide to adopt the corresponding spec change, then yeah.

axic:
Unblocked by the spec.

PR #112, opened 2018-01-23
Add set_option to control fallback (default false)

axic:


src/hera.cpp
```diff
@@ -42,6 +42,10 @@ using namespace std;
 using namespace wasm;
 using namespace HeraVM;
 
+struct hera_instance : evm_instance {
+  bool fallback = false;
```
axic: Does this mean that `fallback` is a member **after** the members of `evm_instance`?

src/hera.cpp
```diff
@@ -231,15 +236,29 @@ static void evm_destroy(struct evm_instance* instance)
...
     .destroy = evm_destroy,
     .execute = evm_execute,
-    .set_option = (evm_set_option_fn)NULL
+    .set_option = evm_set_option
   };
-  struct evm_instance* instance = (struct evm_instance*)calloc(1, sizeof(struct evm_instance));
+  // NOTE: allocating `hera_instance` size here
+  struct evm_instance* instance = (struct evm_instance*)calloc(1, sizeof(struct hera_instance));
   if (instance)
     memcpy(instance, &init, sizeof(init));
```
axic: @chfast any suggestions to clean up the mess here in an idiomatic way?

chfast: Yep, move the initialization of evm_instance to constructor of hera_instance. As in https://github.com/ethereum/evmjit/blob/develop/libevmjit/JIT.cpp#L553-L554.

src/hera.cpp
```diff
@@ -44,6 +44,8 @@ using namespace HeraVM;
 
 struct hera_instance : evm_instance {
   bool fallback = false;
+
+  hera_instance() : abi_version(EVM_ABI_VERSION) {}
```
axic: @chfast `abi_version` being const is super annoying

src/hera.cpp
```diff
@@ -225,24 +232,32 @@ static struct evm_result evm_execute(
...
-    .abi_version = EVM_ABI_VERSION,
-    .destroy = evm_destroy,
-    .execute = evm_execute,
-    .set_option = (evm_set_option_fn)NULL
-  };
-  struct evm_instance* instance = (struct evm_instance*)calloc(1, sizeof(struct evm_instance));
-  if (instance)
-    memcpy(instance, &init, sizeof(init));
-  return instance;
+  struct hera_instance* instance = new struct hera_instance;
```
axic: There's another PR which removes all the `struct` prefixes :)

axic:
@chfast can you please review this again?

src/hera.cpp
```diff
@@ -42,6 +42,12 @@ using namespace std;
 using namespace wasm;
 using namespace HeraVM;
 
+struct hera_instance : evm_instance {
+  bool fallback = false;
+
+  hera_instance() : evm_instance({EVM_ABI_VERSION, NULL, NULL, NULL}) {}
```
chfast: More C++ish would be `nullptr`. But anyway, why not set the functions here? I.e. `{EVM_ABI_VERSION, evm_destroy, evm_execute, evm_set_option}`.

Also, I would name the function with `hera_` prefix, not `evm_` prefix...

axic: > why not set the functions here?

For that I'll need to move code around. And need to add the types as arguments.

> Also, I would name the function with hera_ prefix, not evm_ prefix...

Yeah we could do that too.

axic: I'm merging this now, can you create a PR to clean it up even more? (Needs to be merged for the fallback false support)

PR #113, opened 2018-01-23
Helper functions for overflow checking

jakelang:
To be used in gas charges

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -781,4 +781,18 @@ namespace HeraVM {
     }
     return false;
   }
+  size_t EthereumInterface::highestBitPosition(uint32_t value)
+  {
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -781,4 +781,18 @@ namespace HeraVM {
     }
     return false;
   }
+  size_t EthereumInterface::highestBitPosition(uint32_t value)
```
axic: Where is this needed?

Also, one could use `ffs(value)` (find first set) or `32 - clz(value)` (count leading zeroes) to get the same. These in many CPUs have their own instructions.

jakelang: oh nice,
nvm then

PR #117, opened 2018-01-24
Run Javascript evm2wasm

axic:
Depends on #40.


src/hera.cpp
```diff
@@ -50,6 +56,42 @@ struct hera_instance : evm_instance {
...
+#if HERA_EVM2WASM_JS
+// NOTE: assumes that pattern doesn't contain any formatting characters (e.g. %)
+string mktemp_string(string pattern) {
+  char tmp[PATH_MAX] = { 0 };
+  size_t len = snprintf(tmp, sizeof(tmp), "%s", pattern.c_str());
+  mktemp(tmp);
+  return string(tmp, len);
+}
+
+string evm2wasm(string const& input) {
```
axic: Need to add a comment:
a) expects and outputs a hex string
b) expects and outputs binary

src/hera.cpp
```diff
@@ -50,6 +56,42 @@ struct hera_instance : evm_instance {
...
+  string fileEVM = mktemp_string("/tmp/hera.evm2wasm.evm.XXXXXXXX");
+  string fileWASM = mktemp_string("/tmp/hera.evm2wasm.wasm.XXXXXXXX");
+
+  ofstream os;
+  os.open(fileEVM);
+  os << input;
```
axic: `input` is bytes here, should be translated into a hex string

axic:
@chfast can you review this?

src/hera.cpp
```diff
@@ -50,6 +52,49 @@ struct hera_instance : evm_instance {
 
 namespace {
 
+#if HERA_EVM2WASM_JS
+// NOTE: assumes that pattern doesn't contain any formatting characters (e.g. %)
+string mktemp_string(string pattern) {
+  char tmp[PATH_MAX] = { 0 };
```
chfast: You don't need this tmp on stack. Just resize string to `PATH_MAX`, then check the length of path with `strlen` and shrink the string back.

src/hera.cpp
```diff
@@ -50,6 +52,49 @@ struct hera_instance : evm_instance {
...
+  os.open(fileEVM);
+  // print as a hex sting
+  os << hex;
+  for (uint8_t byte: input)
+    os << static_cast<int>(byte);
+  os.close();
+
+  string cmd = string("evm2wasm.js ") + "-e " + fileEVM + " -o " + fileWASM;
+  if (system(cmd.c_str()) != 0) {
+    unlink(fileEVM.c_str());
```
chfast: If would be nice to refactor it so `unlink()` is in single place.

src/hera.cpp
```diff
@@ -50,6 +52,49 @@ struct hera_instance : evm_instance {
...
+  ofstream os;
+  os.open(fileEVM);
+  // print as a hex sting
+  os << hex;
+  for (uint8_t byte: input)
+    os << static_cast<int>(byte);
+  os.close();
+
+  string cmd = string("evm2wasm.js ") + "-e " + fileEVM + " -o " + fileWASM;
+  if (system(cmd.c_str()) != 0) {
```
chfast: Handle the error code that corresponds to "file not found" separately.

src/hera.cpp
axic: Not sure I follow. `mktemp` modified the template in place. I guess I could:
```
string ret(pattern);
mktemp(ret.c_str());
return ret;
```

src/hera.cpp
axic: Like?
```
int ret = system(...);
unlink(...);
unlink(...);
if (ret != 0) return..
```

src/hera.cpp
```diff
@@ -50,6 +52,49 @@ struct hera_instance : evm_instance {
 
 namespace {
 
+#if HERA_EVM2WASM_JS
+// NOTE: assumes that pattern doesn't contain any formatting characters (e.g. %)
+string mktemp_string(string pattern) {
+  char tmp[PATH_MAX] = { 0 };
+  size_t len = snprintf(tmp, sizeof(tmp), "%s", pattern.c_str());
+  mktemp(tmp);
```
chfast: This will not work.
1. The mktemp expects exactly 6 XXXXXX.
2. mktemp() docs suggest to use mkstemp() http://man7.org/linux/man-pages/man3/mkstemp.3.html.
3. You have to check the returned value.

src/hera.cpp
chfast: You have to use `ret.data()`, not `ret.c_str()`. In C++11 the `.data()` will have the null terminator.

src/hera.cpp

src/hera.cpp
axic: Yeah, whichever has the null terminator.

Is that the preferred way then?

src/hera.cpp
axic: I am not sure, my man page says:
> the number of unique file names mktemp() can return depends on the number of `Xs' provided;

axic: Seems like on BSD it is flexible, but Linux is more picky:
> The last six characters of template must be "XXXXXX" and these are replaced with a string that makes the filename unique.

chfast: This should be fine then:
```
string makeTempPath(string pattern) {
  mktemp(pattern.data());
  return pattern;
}
```

src/hera.cpp
chfast: The `.c_str()` returns const pointer, not useful here.

src/hera.cpp
axic: True, just make `pattern` pass by value.

src/hera.cpp
axic: > 
Accessing the value at data()+size() produces undefined behavior: There are no guarantees that a null character terminates the character sequence pointed by the value returned by this function. See string::c_str for a function that provides such guarantee.

chfast: > Returns a pointer to an array that contains a null-terminated sequence of characters (i.e., a C-string) representing the current value of the string object.
>
>This array includes the same sequence of characters that make up the value of the string object plus an additional terminating null-character ('\0') at the end.
>
>The pointer returned points to the internal array currently used by the string object to store the characters that conform its value.
>
>Both string::data and string::c_str are synonyms and return the same value.

http://www.cplusplus.com/reference/string/string/data/

axic: Damn didn't see mine is valid for C++98 and yours is C++11. Ok, fine then, <11 was just bad.

axic: Cannot use `.data()` as that is `const` and `mktemp` requires non-const.

src/hera.cpp
```diff
@@ -50,6 +52,49 @@ struct hera_instance : evm_instance {
...
+}
+
+vector<uint8_t> evm2wasm(vector<uint8_t> const& input) {
+  string fileEVM = mktemp_string("/tmp/hera.evm2wasm.evm.XXXXXXXX");
+  string fileWASM = mktemp_string("/tmp/hera.evm2wasm.wasm.XXXXXXXX");
+
+  ofstream os;
+  os.open(fileEVM);
+  // print as a hex sting
+  os << hex;
```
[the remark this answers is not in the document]
axic: Actually `evm2wasm` is broken. In the file it is hex, on the CLI it is hex string. Since in pretty much every case it is represented in a hex string I'd just fix `evm2wasm`.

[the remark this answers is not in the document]
axic: Fixed https://github.com/ewasm/evm2wasm/pull/194

axic:
Replaced by #40.

PR #125, opened 2018-01-28
Fix cpp-ethereum build due to libbinaryen.a not found

GreatAttractor:
When building `cpp-ethereum` on Fedora 27, `libbinaryen.a` is generated in `deps/lib64`, not `deps/lib`.
Same type of problem as there was with `MPIR` (https://github.com/ethereum/cpp-ethereum/issues/4159).

### DIFF
--- a/cmake/ProjectBinaryen.cmake
+++ b/cmake/ProjectBinaryen.cmake
@@ -23,7 +23,7 @@ set(source_dir ${prefix}/src/binaryen)
 set(binary_dir ${prefix}/src/binaryen-build)
 # Use source dir because binaryen only installs single header with C API.
 set(binaryen_include_dir ${source_dir}/src)
-set(binaryen_library ${prefix}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}binaryen${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(binaryen_library ${prefix}/${CMAKE_INSTALL_LIBDIR}/${CMAKE_STATIC_LIBRARY_PREFIX}binaryen${CMAKE_STATIC_LIBRARY_SUFFIX})
 # Include also other static libs needed:
 set(binaryen_other_libraries
     ${binary_dir}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}wasm${CMAKE_STATIC_LIBRARY_SUFFIX}


cmake/ProjectBinaryen.cmake
```diff
@@ -23,7 +23,7 @@ set(source_dir ${prefix}/src/binaryen)
 set(binary_dir ${prefix}/src/binaryen-build)
 # Use source dir because binaryen only installs single header with C API.
 set(binaryen_include_dir ${source_dir}/src)
-set(binaryen_library ${prefix}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}binaryen${CMAKE_STATIC_LIBRARY_SUFFIX})
+set(binaryen_library ${prefix}/${CMAKE_INSTALL_LIBDIR}/${CMAKE_STATIC_LIBRARY_PREFIX}binaryen${CMAKE_STATIC_LIBRARY_SUFFIX})
 # Include also other static libs needed:
 set(binaryen_other_libraries
     ${binary_dir}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}wasm${CMAKE_STATIC_LIBRARY_SUFFIX}
```
axic: I think this problem appears for these too?

GreatAttractor: Do you mean the arguments to `set(binaryen_other_libraries`? I'm not sure, but at least for `cpp-ethereum` (with HERA disabled or enabled) my change suffices.

axic: I think all of the `/lib/` pieces should be replaced, shouldn't they?

chfast: Yes.

GreatAttractor:
After trying it I got:
```
make[2]: *** No rule to make target 'deps/src/binaryen-build/lib64/libwasm.a', needed by 'test/testeth'.  Stop.
```

Note that `binaryen_other_libraries` (including `libwasm.a`) do in fact appear in `deps/src/binaryen-build/lib`. This is apparently determined by `CMakeLists.txt` in the `binaryen` sources downloaded during build:

```
# Force output to bin/ and lib/. This is to suppress CMake multigenerator output paths and avoid bin/Debug, bin/Release/ and so on, which is CMake default.
```

chfast:
As mentioned, all occurrences of "/lib/" must be replaced. I'm on it.

chfast:
Ok, let me check out the binaryen CMake script.

PR #127, opened 2018-01-29
[WIP] Fix CREATE control flow

### LINKED ISSUE #122: CREATE doesn't work
See this branch with the test case in order to reproduce: https://github.com/ewasm/tests/tree/create-test

This is all the output I get:

```
Executing...
useGas 300
create0 14 0 132
```

jakelang:
Potentially fixes #122

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -586,10 +586,12 @@ namespace HeraVM {
 
       evm_result create_result;
       context->fn_table->call(&create_result, context, &create_message);
+
       if (create_result.status_code == EVM_SUCCESS) {
         storeUint160(create_result.create_address, resultOffset);
-        lastReturnData.clear();
-      } else if (create_result.output_data) {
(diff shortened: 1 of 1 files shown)


axic:
I think this is wrong.

chfast:
Yes, this assumes that the output from CREATE will be empty in case of SUCCESS (what is true). But you have `.clear()` in single place.

axic:
Clear is applied in both cases currently. What is the actual bug this solves?

jakelang:
I looked at it again, this probably doesnt solve the bug but does clean up the code a bit. assuming the client works correctly this should be exactly the same

axic:
Well there's no guarantee on EVM-C for this atm.

jakelang:
That's why it would be helpful to separate the conditionals for output data and return code, so that it is easier to tell if there is something wrong.

PR #128, opened 2018-01-31
Add raw buffer access to the instance memory

axic:


src/shell-interface.h
```diff
@@ -50,6 +50,10 @@ struct ShellExternalInterface : ModuleInstance::ExternalInterface {
 
    public:
     Memory() {}
+    char* rawbuffer(unsigned offset, unsigned length) {
+      assert(size() <= (offset + length), "memory: requested raw buffer is too short");
```
jakelang: Do we want to use heraAssert macro here, or settle for plain assert?

axic: It is a different file where `heraAssert` is not available.

jakelang: Should it not be moved to a place where it is more accessible, then?
I think that unless heraAssert is specifically intended for EEI, it should be put in a header of project-wide visibility so that any objects that use it do not have to include eei.h.

axic:
This PR, when based atop #172, could change the `copy_code` call pointing to an actual buffer instead of a temporary vector.

chfast:
Still WIP?

src/shell-interface.h
axic: After #187 it can use those exceptions.

jakelang:
rebased.

axic:
I think with the rebase some changes were lost? It was changing `eei.cpp` for `extcodecopy` and also using `heraAssert`.

jakelang:
@axic trying to figure out why rebasing would cause this. Maybe just move your local copy of the branch here again?

axic:
I don't have a local copy. I just delete my copy after push to avoid stale branches.

I'll just recreate it.

axic:
@jakelang recovered the commit hash from circleci logs and hence could push it again :)

src/shell-interface.h
```diff
@@ -50,6 +51,10 @@ struct ShellExternalInterface : ModuleInstance::ExternalInterface {
 
    public:
     Memory() {}
+    char* rawbuffer(unsigned offset, unsigned length) {
```
chfast: I'm pretty sure now that using `uint8_t*` type for bytes in EVMC was good idea.

axic: This is `char*` because the underlying type is `char`.

src/shell-interface.h
```diff
@@ -50,6 +51,10 @@ struct ShellExternalInterface : ModuleInstance::ExternalInterface {
...
+      heraAssert(size() <= (offset + length), "Memory is shorter than requested segment"); // "memory: requested raw buffer is too short"
```
axic: This should be `>=`.

jakelang:
Is this blocked by something or not needed?

src/eei.h
```diff
@@ -81,6 +81,10 @@ struct EthereumInterface : ShellExternalInterface {
   inline size_t memorySize() const { return memory.size(); }
   inline void memorySet(size_t offset, uint8_t value) { memory.set<uint8_t>(offset, value); }
   inline uint8_t memoryGet(size_t offset) { return memory.get<uint8_t>(offset); }
+  inline uint8_t* memoryPointer(size_t offset, size_t length) {
```
axic: Maybe `memorySlicePointer`?

axic:
Although we haven't profiled Hera, I could imagine memory operations can be a big overhead. Maybe we need to get this one merged. What do you think @chfast @cdetrio?

axic:
Could be done for wabt too. This is its memory definition:
```
struct Memory {
  Memory() = default;
  explicit Memory(const Limits& limits)
      : page_limits(limits), data(limits.initial * WABT_PAGE_SIZE) {}

  Limits page_limits;
  std::vector<char> data;
};
```

src/wabt.cpp
```diff
@@ -61,6 +61,10 @@ class WabtEthereumInterface : public EthereumInterface {
   size_t memorySize() const override { return m_wasmMemory->data.size(); }
   void memorySet(size_t offset, uint8_t value) override { m_wasmMemory->data[offset] = static_cast<char>(value); }
   uint8_t memoryGet(size_t offset) override { return static_cast<uint8_t>(m_wasmMemory->data[offset]); }
+  uint8_t* memoryPointer(size_t offset, size_t length) override {
+    ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Memory is shorter than requested segment");
+    return reinterpret_cast<uint8_t*>(m_wasmMemory->data.data() + offset);
```
chfast:
```suggestion
    return reinterpret_cast<uint8_t*>(&m_wasmMemory->data[offset]);
```

src/wavm.cpp
```diff
@@ -63,6 +63,10 @@ class WavmEthereumInterface : public EthereumInterface {
   size_t memorySize() const override { return Runtime::getMemoryNumPages(m_wasmMemory) * 65536; }
   void memorySet(size_t offset, uint8_t value) override { (Runtime::memoryArrayPtr<U8>(m_wasmMemory, offset, 1))[0] = value; }
   uint8_t memoryGet(size_t offset) override { return (Runtime::memoryArrayPtr<U8>(m_wasmMemory, offset, 1))[0]; }
+  uint8_t* memoryPointer(size_t offset, size_t length) override {
+    ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Memory is shorter than requested segment");
+    return reinterpret_cast<uint8_t*>(m_wasmMemory + offset);
```
chfast:
```suggestion
    return reinterpret_cast<uint8_t*>(&m_wasmMemory[offset]);
```

src/wabt.cpp
```diff
@@ -61,6 +61,10 @@ class WabtEthereumInterface : public EthereumInterface {
   size_t memorySize() const override { return m_wasmMemory->data.size(); }
   void memorySet(size_t offset, uint8_t value) override { m_wasmMemory->data[offset] = static_cast<char>(value); }
   uint8_t memoryGet(size_t offset) override { return static_cast<uint8_t>(m_wasmMemory->data[offset]); }
+  uint8_t* memoryPointer(size_t offset, size_t length) override {
+    ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Memory is shorter than requested segment");
```
chfast: I'm not sure the "shorter" is the right word.

src/wavm.cpp
```diff
@@ -63,6 +63,10 @@ class WavmEthereumInterface : public EthereumInterface {
   size_t memorySize() const override { return Runtime::getMemoryNumPages(m_wasmMemory) * 65536; }
   void memorySet(size_t offset, uint8_t value) override { (Runtime::memoryArrayPtr<U8>(m_wasmMemory, offset, 1))[0] = value; }
   uint8_t memoryGet(size_t offset) override { return (Runtime::memoryArrayPtr<U8>(m_wasmMemory, offset, 1))[0]; }
+  uint8_t* memoryPointer(size_t offset, size_t length) override {
+    ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Memory is shorter than requested segment");
+    return Runtime::memoryArrayPtr<U8>(m_wasmMemory, offset, length);
```
axic: Based on https://github.com/WAVM/WAVM/blob/fa5434e03efbc2154ecf4aafede169da76a4da40/Include/Runtime/Runtime.h#L301

src/wabt.cpp
chfast: Can you change it to "Memory (size) is smaller ..."?

axic: Did find this instance in Boost where they the same sentence :wink::
>   * @param n The number of bytes to return in the buffer sequence. If the
>   * underlying memory is shorter, the buffer sequence represents as many bytes
>   * as are available.

I can change it if it is important.

PR #134, opened 2018-02-18
run ewasm tests on circleCI

cdetrio:
run ewasm tests and use circle cache for cpp-eth build.


axic:
Great work! This prebuilt cpp-eth caching stuff looks a bit unusual, but seemingly it works.

axic:
Probably we need to change this over to the `testeth` repo as opposed to `cpp-ethereum`.

Issue #137, opened 2018-03-01
`return` doesn't account for data segments


jwasinger:
Here is the corresponding test case for this issue: https://github.com/ewasm/tests/pull/10

Error:
`InternalError: Out of bounds (destination) memory copy.`

hugo-dc:
I got the same error while testing callDelegate: https://github.com/ewasm/hera/issues/129#issuecomment-363639790

Issue #129, opened 2018-02-06
callDelegate result is 0 (successful) when it runs Out of Gas


hugo-dc:
callDelegate result is 0 (successful) when it runs Out of Gas.

I created this test case for this: 
https://github.com/hugo-dc/tests/blob/fe9fa523316f00eadb7f4620d7e7376d5532289e/src/GeneralStateTestsFiller/stEWASMTests/utils/callDelegateOOGFiller.yml

axic:
Can you try doing a `revert` within a delegatecall?

hugo-dc:
Same result for `revert`, there is an Internal error while calling `revert` or `result`:

```
InternalError: Out of bounds (destination) memory copy.
```
It is thrown here: https://github.com/ewasm/hera/blob/master/src/eei.cpp#L655

`dst.size() = 0`, `length = 2`

Not sure if this problem is new, or if @lrettig already had this problem when he wrote the `revert` test case: https://github.com/ewasm/tests/blob/wasm-tests/src/GeneralStateTestsFiller/stEWASMTests/utils/revertFiller.yml (I ran this test case and got the same InternalError)

lrettig:
I don't recall seeing this error, let me know if I can help take a look!

axic:
Is this still a problem?

hugo-dc:
It is not throwing the InternalError anymore (InternalError: Out of bounds (destination) memory copy.), but I still receive 0 (success) as a result.

hugo-dc:
Hera is catching the OOG exception correctly, and it is setting status 2 here: https://github.com/ewasm/hera/blob/master/src/hera.cpp#L323 

But, after this line is executed https://github.com/ewasm/hera/blob/master/src/eei.cpp#L644 , i see the status code as 0 (success), i also can see the result is set to success here https://github.com/ethereum/cpp-ethereum/blob/develop/libevm/ExtVMFace.cpp#L227

axic:
@chfast can you have a quick look into this?

chfast:
Can you give me full command to reproduce it?

chfast:
I've found the cause of the problem. Will fix it soon.

axic:
@hugo-dc can you retest with latest cpp-ethereum?

hugo-dc:
[...]

PR #138, opened 2018-03-01
Zero-fill return data  buffer correctly 

jwasinger:
(probably) fixes https://github.com/ewasm/hera/issues/137 https://github.com/ewasm/hera/issues/129

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -487,7 +487,7 @@ namespace HeraVM {
 
       HERA_DEBUG << (import->base == Name("revert") ? "revert " : "return ") << hex << offset << " " << size << dec << "\n";
 
-      result.returnValue.clear();
+      result.returnValue = vector<uint8_t>(size);
       loadMemory(offset, result.returnValue, size);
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -432,7 +432,11 @@ namespace HeraVM {
 
       HERA_DEBUG << (import->base == Name("revert") ? "revert " : "return ") << hex << offset << " " << size << dec << "\n";
 
+      //zero-fill the return buffer
       result.returnValue.clear();
+      for (uint32_t i = 0; i < size; i++)
```
axic: This should go until `offset`, but according to @chfast this is broken anyway, we shouldn't clear and just randomly write, but rather properly `resize` which will do the zero filling.

axic: Apparently `resize` doesn't do zero filling if the buffer is truncated.

@jwasinger I don't fully get why you need to populate with zeroes when the same slots are overwritten by `loadMemory`?

Note, `loadMemory` should write to `returnValue[0 ... size]` and read from `memory[offset .. offset+size]`.

PR #139, opened 2018-03-01
allow _main as a contract entrypoint

jwasinger:
I think perhaps we should consider a better name for the contract entrypoint as `main` makes it difficult (impossible?) to compile C programs to ewasm-compliant wasm.  I have suggested `_main` here, but I think it could be changed to something better.

Ideas?

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -150,7 +150,7 @@ void execute(
 
   // This should be caught during deployment time by the Sentinel.
   // TODO: validate for other conditions too?
-  heraAssert(module.getExportOrNull(Name("main")) != nullptr, "Contract entry point (\"main\") missing.");
+  heraAssert(module.getExportOrNull(Name("main")) != nullptr && module.getExportOrNull(Name("_main")) != nullptr, "Contract entry point (\"main\" or \"_main\") missing.");
 
   // NOTE: DO NOT use the optimiser here, it will conflict with metering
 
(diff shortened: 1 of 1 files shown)


axic:
Hint: should check out https://github.com/ewasm/ewasm-cleanup, which intends to translate such code :)

Also I have a binaryen based version almost finished.

jwasinger:
oh nice!

src/hera.cpp
```diff
@@ -150,15 +150,21 @@ void execute(
 
   // This should be caught during deployment time by the Sentinel.
   // TODO: validate for other conditions too?
-  heraAssert(module.getExportOrNull(Name("main")) != nullptr, "Contract entry point (\"main\") missing.");
+  heraAssert(module.getExportOrNull(Name("main")) != nullptr && module.getExportOrNull(Name("_main")) != nullptr, "Contract entry point (\"main\" or \"_main\") missing.");
```
jakelang: ```module.getExportOrNull(Name("main")) != nullptr && module.getExportOrNull(Name("_main")) != nullptr```
The assertion should not require that both "main" and "_main" exports exist. Should be using OR operator instead of AND.

jwasinger: yeah good call.  I had that change in my local environment but forgot to push it up.

axic:
I think I'll close this, we should finish the other tooling to preprocess the files.

PR #140, opened 2018-03-04
Add debug logging for return

jwasinger:
Logs the value that is being returned.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -437,6 +437,14 @@ namespace HeraVM {
 
       result.isRevert = import->base == Name("revert");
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -437,6 +437,14 @@ namespace HeraVM {
 
       result.isRevert = import->base == Name("revert");
 
+#if HERA_DEBUGGING
```
axic: I think this debugging should be in `hera.cpp`.

axic:
@jwasinger do we still need this?

jwasinger:
Nope

PR #146, opened 2018-03-13
Require that all symbols can be resolved at link time. Added missing thread libraries

esteve:


axic:
Can you split the commits?

Also would you be interested figuring out why `pthread` is needed by binaryen and if it could be disabled?

esteve:
> Can you split the commits?

Done.

> Also would you be interested figuring out why pthread is needed by binaryen and if it could be disabled?

Sure, though I need to check if it's binarygen or something else, but my laptop is still building another brnach.

esteve:
> Also would you be interested figuring out why `pthread` is needed by binaryen and if it could be disabled?

It can't be disabled unfortunately:

https://github.com/WebAssembly/binaryen/blob/master/src/support/CMakeLists.txt#L8
https://github.com/WebAssembly/binaryen/blob/master/src/support/threads.cpp
https://github.com/WebAssembly/binaryen/blob/master/src/support/threads.h#L28-L29

src/CMakeLists.txt
```diff
@@ -18,4 +20,11 @@ if(HERA_METERING_CONTRACT)
 endif()
 
 target_include_directories(hera PUBLIC ${CMAKE_CURRENT_SOURCE_DIR} PRIVATE ${evm_include_dir})
-target_link_libraries(hera PRIVATE binaryen::binaryen)
+target_link_libraries(hera PRIVATE binaryen::binaryen Threads::Threads)
+if(NOT WIN32)
+  if(CMAKE_COMPILER_IS_GNUCXX)
+    set_target_properties(hera PROPERTIES LINK_FLAGS "-Wl,--no-undefined")
+  elseif(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    set_target_properties(hera PROPERTIES LINK_FLAGS "-Wl,-undefined,error")
```
chfast: Will that work on Linux?

axic:
It turns out the optimisation passes require it: https://github.com/WebAssembly/binaryen/blob/d52213c3f5e96bb3450721d96aa68d3c5e0865b6/src/passes/pass.cpp#L275

src/CMakeLists.txt
chfast: I did a small research. Both flags seems to be working on Linux. Also in EVMJIT I only have single one: https://github.com/ethereum/evmjit/blob/develop/CMakeLists.txt#L41.

Can you use one for both GCC and clang?

esteve: The second line is for OSX, which uses Clang's linker, instead of GCC's `binutils`

chfast: Let me elaborate a bit:

1. I did build it on Linux with clang. The option `-Wl,-undefined,error` was used. I don't have clang's linker installed, only ld and/or gold. I believe ld understands `-undefined,error`.

2. In EVMJIT, I'm using only `-Wl,--no-undefined` variant and it builds fine on macOS (although I have not checked for some time).

It might be better to setup CI for macOS to resolve it. https://github.com/ewasm/hera/issues/152

esteve: > I did build it on Linux with clang. The option -Wl,-undefined,error was used. I don't have clang's linker installed, only ld and/or gold. I believe ld understands -undefined,error.

It seems that `ld` just eats up `-Wl,-undefined,error`, but doesn't really do anything. Let's take this [code](https://gist.github.com/esteve/0a237aabc7bf848e60731473c48ccc15#file-linker_flags-cpp) for example, it's missing the symbol for `Foo::bar`:

`ld` correctly reports that it's missing if we pass `-Wl,--no-undefined`:

```sh
$ g++ -Wall -fPIC -o linker_flags -shared linker_flags.cpp -Wl,--no-undefined
/tmp/ccIoK6gQ.o: In function `func()':
linker_flags.cpp:(.text+0x10): undefined reference to `Foo::bar()'
collect2: error: ld returned 1 exit status
```

Whereas it just goes silent when passing `-Wl,-undefined,error`
```sh
$ g++ -Wall -fPIC -o linker_flags -shared linker_flags.cpp -Wl,-undefined,error
```

> In EVMJIT, I'm using only -Wl,--no-undefined variant and it builds fine on macOS (although I have not checked for some time).

This might be because the entire project is built as a single binary, so the missing symbols in hera are solved via a 3rd party, but if we spin off hera as a shared library, it may have unresolved symbols down the road.

PR #147, opened 2018-03-13
Export hera_create() so that it can evenetually be loaded dynamically

esteve:
This is so that hera can be being dynamically loaded in the future.

### DIFF
--- a/src/hera.h
+++ b/src/hera.h
@@ -25,12 +25,24 @@
 #ifndef __HERA_H
 #define __HERA_H
 
+#if defined _MSC_VER || defined __MINGW32__
+# define HERA_EXPORT __declspec(dllexport)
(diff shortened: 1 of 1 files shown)


src/hera.h
```diff
@@ -25,12 +25,24 @@
 #ifndef __HERA_H
 #define __HERA_H
 
+#if defined _MSC_VER || defined __MINGW32__
+# define HERA_EXPORT __declspec(dllexport)
+# define HERA_IMPORT __declspec(dllimport)
+#elif __GNU__ >= 4
+# define HERA_EXPORT __attribute__((visibility("default")))
```
chfast: I think the default visibility is still "default", not "hidden".

esteve: That's correct, I just added it to make it more explicit and clearer.

PR #148, opened 2018-03-13
WASM engine abstraction layer

### LINKED ISSUE #156: Abstract VM / interfaces to support multiple VMs

jakelang:
Fixes #156.

Compatibility layer to enable operation of Hera with multiple VMs. Also CMake targets.

### DIFF
--- a/src/eei.h
+++ b/src/eei.h
@@ -27,44 +27,166 @@ using namespace wasm;
 
 namespace HeraVM {
 
+/*
+ * Execution output data structure. Used in WasmEngine class
+ */
 struct ExecutionResult {
+  ExecutionResult() { }
+  ExecutionResult(uint64_t _gasLeft):
+    gasLeft(_gasLeft)
+  { }
   uint64_t gasLeft = 0;
   std::vector<uint8_t> returnValue;
   bool isRevert = false;
 };
 
-struct EthereumInterface : ShellExternalInterface {
-  EthereumInterface(
-    evmc_context* _context,
+/* 
+ * Base class for EEI implementations.
+ * Children of this class implement WASM engine-specific code for byte memory access
+ *   and passing EEI arguments and returns between the WASM module and the abstract interface.
+ *
+ *   See eei.cpp for example
+ */
+class EEI {
+public:
+  EEI(evmc_context *_context,
     std::vector<uint8_t> const& _code,
+    std::vector<uint8_t> const& _state_code,
     evmc_message const& _msg,
     ExecutionResult & _result,
-    bool _meterGas
-  ):
-    ShellExternalInterface(),
+    bool _meterGas) :
     context(_context),
     code(_code),
+    state_code(_state_code),
     msg(_msg),
     result(_result),
     meterGas(_meterGas)
-  { }
+    { }
 
-  Literal callImport(Import *import, LiteralList& arguments) override;
-#if HERA_DEBUGGING
-  Literal callDebugImport(Import *import, LiteralList& arguments);
-#endif
+protected:
+/*
+ * Gas charge schedule.
+ */
+  struct GasSchedule {
+    static constexpr unsigned storageLoad = 200;
+    static constexpr unsigned storageStoreCreate = 20000;
+    static constexpr unsigned storageStoreChange = 5000;
+    static constexpr unsigned log = 375;
+    static constexpr unsigned logData = 8;
+    static constexpr unsigned logTopic = 375;
+    static constexpr unsigned create = 32000;
+    static constexpr unsigned call = 700;
+    static constexpr unsigned copy = 3;
+    static constexpr unsigned blockhash = 800;
+    static constexpr unsigned balance = 400;
+    static constexpr unsigned base = 2;
+    static constexpr unsigned verylow = 3;
+    static constexpr unsigned extcode = 700;
+    static constexpr unsigned selfdestruct = 5000;
+    static constexpr unsigned valuetransfer = 9000;
+    static constexpr unsigned callNewAccount = 25000;
+  };
 
-  void importGlobals(std::map<Name, Literal>& globals, Module& wasm) override;
+/*
+ * EEI host functions.
+ * WASM-engine specific code is only responsible for validating imports,
+ * passing arguments to these functions, and passing their return values.
+ *
+ * See https://github.com/ewasm/design/blob/master/eth_interface.md
+ */
+  void eth_useGas(uint64_t gas);
+  uint64_t eth_getGasLeft();
 
-  void trap(const char* why) override {
-    // TODO: use a specific error code here?
-    ensureCondition(false, OutOfGas, std::string("Trap condition: ") + why);
-  }
+  void eth_getAddress(uint32_t resultOffset);
+  void eth_getBalance(uint32_t addressOffset, uint32_t resultOffset);
+  void eth_getBlockHash(int64_t number, uint32_t resultOffset);
 
-private:
-  void takeGas(uint64_t gas);
+  virtual void trap(const char* why) = 0;
+
+  uint32_t eth_call(
+    int64_t gas, 
+    uint32_t addressOffset,
+    uint32_t valueOffset,
+    uint32_t dataOffset,
+    uint32_t dataLength);
+  uint32_t eth_callCode(
+    int64_t gas,
+    uint32_t addressOffset,
+    uint32_t valueOffset,
+    uint32_t dataOffset,
+    uint32_t dataLength);
+  uint32_t eth_callDelegate(
+    int64_t gas,
+    uint32_t addressOffset,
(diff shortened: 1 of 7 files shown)


src/hera.h
```diff
@@ -33,6 +33,16 @@ struct evm_instance;
 
 struct evm_instance* hera_create(void);
 
+typedef enum wasm_vm {
+#if WABT_SUPPORTED
```
axic: This shouldn't have ifdefs.

jakelang: this is for CMake build option. Why not?

axic: What is the reason ifdef enums?

jakelang: hera_instance contains one such enum, and preprocessor also controls which options are available with evm_set_option.

axic: I mean the ifdef only needs to be present in `evm_set_option`, there wouldn't be any other way to select them?

jakelang: I don't think it makes a difference at all, but why define constants for VMs that are not built in?

src/hera.cpp
```diff
@@ -263,17 +265,14 @@ evm_result evm_execute(
   evm_result ret;
   memset(&ret, 0, sizeof(evm_result));
 
-  try {
-    heraAssert(msg->gas >= 0, "Negative startgas?");
-
-    ExecutionResult result;
-    result.gasLeft = (uint64_t)msg->gas;
+  hera_instance* hera = static_cast<hera_instance*>(instance);
```
axic: These changes seem not be needed?

src/hera.cpp
```diff
@@ -357,6 +359,20 @@ int evm_set_option(
     return 1;
   }
 #endif
+  if (strcmp(name, "vm") == 0) {
+    if (strcmp(value, "binaryen") == 0)
```
axic: Could use a `map<string,hera_vm>` and that would simplify this a lot.

src/hera.h
```diff
@@ -45,6 +45,16 @@ struct evm_instance;
...
+  WABT,
+#endif
+#if WAVM_SUPPORTED
+  WAVM,
+#endif
+  BINARYEN
```
axic: I'd prefer this to have a prefix, such as `VM_`.

src/vm.h
```diff
@@ -0,0 +1,72 @@
...
+#include "eei.h" /* for ExecutionResult */
+
+#include <evm.h>
+
+using namespace std;
+using namespace HeraVM;
+
+class WasmVM
+{
+private:
```
axic: Please follow the style of eei.h. Public constructor, public methods, private methods, private members.

src/vm.h
```diff
@@ -0,0 +1,72 @@
...
+	           vector<uint8_t> const& code, 
+		   evm_message const& msg,
+		   evm_context *context)
+	{
+		this->vm = vm;
+		this->msg = msg;
+		this->output.gasLeft = (uint64_t)msg.gas;
+		this->context = context;
+		for (size_t i = 0; i < code.size(); ++i)
+			this->code.push_back(code[i]);
```
axic: I don't think there's a need for this, a simple assignment will make a copy.

src/vm.h
```diff
@@ -0,0 +1,72 @@
...
+	int runWavm();
+	#endif
+
+public:
+	WasmVM(wasm_vm const vm, 
+	           vector<uint8_t> const& code, 
+		   evm_message const& msg,
+		   evm_context *context)
+	{
+		this->vm = vm;
```
axic: I'd prefer the way it is done in `eei.h`, have the private members with a prefix `m_`, the arguments have an underscore prefix and it can be assigned simply:

```
WasmVM(vm, code,...) : m_vm(vm), m_code(code), ... {}
```

src/hera.cpp
jakelang: this was moved from line 276 so that the WasmVM object can be instantiated with the correct options, without needing a redundant declaration of hera_instance.

src/vm.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+int WasmVM::runBinaryen()
+{
+	Module module;
+
+	try {
+		WasmBinaryBuilder parser(module, reinterpret_cast<vector<char> const&>(this->code), false);
+		parser.read();
+	} catch (ParseException &p) {
+		return 1;
```
axic: It would be nice not to loose the exception information we had here. Just like it is done currently.

src/vm.h
```diff
@@ -0,0 +1,72 @@
...
+	vector<uint8_t> code;
+	evm_message msg;
+	evm_context *context;
+
+	int runBinaryen();
+	#if WABT_SUPPORTED
+	int runWabt();
+	#endif
+	#if WAVM_SUPPORTED
+	int runWavm();
```
axic: If this is C++, each of those VMs implement their own class with a single `execute()` method.

src/vm.h
jakelang: private members in eei.h do not have the m_ prefix.

src/vm.h
jakelang: Is it a bad idea to leave the execution result as a public member?

src/hera.cpp
jakelang: Wouldn't populating the map be just as tedious?

jakelang:
@axic can you review the recent changes here?

jakelang:
@axic this is updated for latest master now. Please review when you have time, we should get this merged at some point

axic:
Rebased.

src/eei.h
```diff
@@ -31,27 +31,71 @@
 #include "hera.h"
 #include "exceptions.h"
 
+#define heraAssert(condition, msg) { \
```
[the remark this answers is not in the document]
jakelang: fixed

jakelang:
@axic maybe it is wise to merge this, so as to enable working on other features such as the debugging much more cleanly. Full EEI abstraction can be included in a separate PR, keeping things a bit more organized.

src/vm.cpp
```diff
@@ -32,9 +32,12 @@ int BinaryenVM::execute()
 {
   Module module;
   
+  //Copy over bytecode instead of type-punning to avoid breaking alias rules
+  vector<char> const bytecode(code.begin(), code.end());
```
axic: What alias rules exactly?

src/hera.cpp
```diff
@@ -394,18 +302,41 @@ evmc_result hera_execute(
       ensureCondition(_code.size() > 5, ContractValidationFailure, "Invalid contract or metering failed.");
     }
 
-    execute(context, _code, state_code, *msg, result, meterInterfaceGas);
+    /*
+     * Instantiate, execute, and extract results from the correct WASM engine
+     */
+    ExecutionResult vmresult;
```
axic: Why change this from `result`?

src/hera.cpp
```diff
@@ -394,18 +302,41 @@ evmc_result hera_execute(
...
+      vm.execute();
+      vmresult = vm.getResult();
+    #endif
+    #if WAVM_SUPPORTED
+    case VM_WAVM:
+      WavmVM vm = WavmVM(_code, _state_code, *msg, context, meterInterfaceGas);
+      vm.execute();
+      vmresult = vm.getResult();
+    #endif
+    default:
```
axic: I don't think this should have a default case (or at least the default case should have an assert). Rather only specific ones to the VMs.

src/hera.cpp
jakelang: naming-wise it's a bit more clear IMO.

src/vm.cpp
jakelang: The compiler does not allow me to reinterpret_cast vector<uint8_t> to vector<char> anymore.

src/hera.cpp
jakelang: `ExecutionResult executionResult;` would be confusing i think.

src/hera.cpp
jakelang: I will fix that

jakelang:
Rebased.

src/vm.h
```diff
@@ -0,0 +1,139 @@
...
+    output(_msg.gas),
+    meterGas(_meterGas)
+  { }
+ 
+ /* 
+  * Pure virtual method implemented by child classes of WasmEngine
+  * Calling this should trigger execution of the wasm code provided in the constructor
+  *   and instantiation of an EEI object.
+  */
+  virtual int execute() = 0;
```
axic: I think we should go with having all those parameters in the `execute` method. But for the best design we need to prototype wabt and wavm integration (with a few eei methods).

axic:
@jakelang can we close this now or is there something not merge from this?

jakelang:
@axic I think everything relevant got merged. Closing it

PR #149, opened 2018-03-13
CMake: Fix missing include

chfast:


chfast:
It makes CMAKE_INSTALL_LIBDIR to have any value. I don't know how did you build it without it...

PR #151, opened 2018-03-13
Add .clang-format file

### LINKED ISSUE #150: Setup clang-format
It's time to do it while the code is still small and ugly anyway.

jakelang:
Fixes #150

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,41 @@
+---
+Language:        Cpp
+BasedOnStyle:  Chromium
+AccessModifierOffset: -4
+AlignAfterOpenBracket: DontAlign
+BinPackParameters: true
+BraceWrapping:
+  AfterClass:      true
+  AfterControlStatement: true
+  AfterEnum:       true
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: true
+  AfterStruct:     true
+  AfterUnion:      true
+  BeforeCatch:     true
+  BeforeElse:      true
+  SplitEmptyFunction: false
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: false
+ColumnLimit:     100
+ConstructorInitializerIndentWidth: 2
+IncludeCategories:
+  - Regex:           '^".*'
+    Priority:        1
+  - Regex:           '^<boost.*'
+    Priority:        98
+  - Regex:           '^<.*\.h>'
+    Priority:        2
+  - Regex:           '^<.*'
+    Priority:        99
+  - Regex:           '.*'
+    Priority:        4
+IncludeIsMainRegex: '(Test)?$'
+IndentCaseLabels: false
+IndentWidth:     4
+MaxEmptyLinesToKeep: 2
+PenaltyBreakAssignment: 1
+PenaltyBreakComment: 50
+TabWidth:        4
+...
--- a/.clang-format
+++ b/.clang-format
@@ -1,5 +1,6 @@
 ---
 Language:        Cpp
+Standard:        Cpp11
 BasedOnStyle:  Chromium
 AccessModifierOffset: -2
 AlignAfterOpenBracket: DontAlign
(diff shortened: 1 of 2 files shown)


chfast:
Don't ever do this.

jakelang:
@chfast ?
was working on fixing it

chfast:
You can fix it, but reformatting all the code will never be accepted. You should propose .clang-format config first. I recommend the one from ethminer.

chfast:
Sorry I closed it prematurely.

jakelang:
ah ok. will roll back then. This format file is from cpp-ethereum

chfast:
About fixing it: clang-format reorders the includes (if asked). This can be the reason for build errors.

jakelang:
Apparently this was not the issue, as I had tried earlier to simply switch around the only two includes.

jakelang:
This has not been touched in a long time. Do we still want to add a format file?

chfast:
I'm really for it. It can be different style, but it would make my life easier if I could reformat my changes with a single click.

jakelang:
What style would you prefer?

chfast:
Me? This one: https://github.com/ethereum-mining/ethminer/blob/master/.clang-format.

jakelang:
Alright, I am using the ethminer clang-format file. I have no preference for the style outside of plain C.

axic:
Seems like the webkit style is the closest: https://webkit.org/code-style-guidelines/

Need to check these out:
> BasedOnStyle (string)
> The style used for all options not specifically set in the configuration.
> 
> This option is supported only in the clang-format configuration (both within -style='{...}' and the .clang-format file).
> 
> Possible values:
> 
> LLVM A style complying with the LLVM coding standards
> Google A style complying with Google’s C++ style guide
> Chromium A style complying with Chromium’s style guide
> Mozilla A style complying with Mozilla’s style guide
> WebKit A style complying with WebKit’s style guide

chfast:
If you want to experiment with it,

- make sure you have the latest version of clang-format (6),
- dump the config of the style you thinks is the closest: 
```
clang-format-6.0 -style=WebKit -dump-config > .clang-format
```
. You will be able to see all available options.

axic:
Yeah I am reading the documentation and have a pretty good version now.

chfast:
> I do not really like formatting changes like this:
> 
> -bool hasWasmPreamble(vector<uint8_t> const& _input) {
> -  return
> -    _input.size() >= 8 &&
> -    _input[0] == 0 &&
> -    _input[1] == 'a' &&
> -    _input[2] == 's' &&
> -    _input[3] == 'm' &&
> -    _input[4] == 1 &&
> -    _input[5] == 0 &&
> -    _input[6] == 0 &&
> -    _input[7] == 0;
> +namespace
> +{
> +bool hasWasmPreamble(vector<uint8_t> const& _input)
> +{
> +    return _input.size() >= 8 && _input[0] == 0 && _input[1] == 'a' && _input[2] == 's' &&
> +           _input[3] == 'm' && _input[4] == 1 && _input[5] == 0 && _input[6] == 0 && _input[7] == 0;
>  }

I don't think there is any option to change how the expressions are packed. If you want hand-crafted expressions, wrap them with `// clang-format off`.

chfast:
Can I use it for new code?

jakelang:
Rebased.

axic:
@chfast please note I'll adjust the settings, just wanted to push CI first.

chfast:
I've noticed you touched this PR, so I dumped the style I'm currently using for Hera. I don't care which one is used in the end, but having one is big improvement for my daily workflow.

chfast:
What's the progress?

chfast:
I will take a look next week. I've notices clang-format has problems with formatting arrays of lambdas.

You don't want to format CMake files...

axic:
> You don't want to format CMake files...

Yeah the final version of this PR won't do that.

axic:
Should copy the rules from https://github.com/wasmx/fizzy.

PR #154, opened 2018-03-14
 Fix ifdef for evm2wasm.js EVM-C option

axic:
Fixes #40.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -351,7 +351,7 @@ int evm_set_option(
     hera->fallback = strcmp(value, "true") == 0;
     return 1;
   }
-#if EVM2WASM
+#if HERA_EVM2WASM
   if (strcmp(name, "evm2wasm.js") == 0) {
     hera->use_evm2wasm_js = strcmp(value, "true") == 0;
     return 1;


src/hera.cpp
```diff
@@ -134,7 +134,12 @@ vector<uint8_t> evm2wasm_js(vector<uint8_t> const& input) {
   string fileEVM = mktemp_string("/tmp/hera.evm2wasm.evm.XXXXXX");
   string fileWASM = mktemp_string("/tmp/hera.evm2wasm.wasm.XXXXXX");
 
-  if (fileEVM.size() == 0 || fileWASM.size() == 0)
+  if (
+    fileEVM.size() == 0 ||
```
chfast: `.size() == 0` <=> `.empty()`.

src/hera.cpp
```diff
@@ -134,7 +134,12 @@ vector<uint8_t> evm2wasm_js(vector<uint8_t> const& input) {
...
+    fileWASM.size() == 0 ||
+    unlink(fileEVM.data()) != 0 ||
```
chfast: Where it was moved from?

I think this is wrong. If removing of the first file fails, the removal of the second one is not even tried.

axic: It is a failure case anyway?

PR #155, opened 2018-03-14
 Run metering on CREATE output only if it returned wasm bytecode

axic:


axic:
@hugo-dc this should fix your issue

src/hera.cpp
```diff
@@ -56,6 +56,10 @@ struct hera_instance : evm_instance {
 
 namespace {
 
+bool hasWasmSignature(vector<uint8_t> const& _input) {
```
chfast: Not sure, but maybe "magic number" would be a better name instead of "signature". Or "prefix". Or `isWasmCode()`.

axic: It is called signature in the spec IIRC.

axic: According to the spec, we should extend that to 8 bytes and called is preamble. The version byte is little endian 32 bits, so the next 3 bytes would be zero.

PR #160, opened 2018-03-14
Fix issue with EVM bytecode mangling

jwasinger:
When running transcompiled EVM code (from evm2wasm.js), individual bytes are not correctly padded when storing to a temp file before transcompilation.  The result is that the bytecode is mangled.

This PR fixes the issue.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -153,7 +153,7 @@ vector<uint8_t> evm2wasm_js(vector<uint8_t> const& input) {
   ofstream os;
   os.open(fileEVM);
   // print as a hex sting
-  os << hex;
+  os << setfill('0') << setw(2) << hex;
   for (uint8_t byte: input)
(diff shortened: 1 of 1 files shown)


src/hera.cpp
```diff
@@ -154,8 +154,12 @@ vector<uint8_t> evm2wasm_js(vector<uint8_t> const& input) {
   os.open(fileEVM);
   // print as a hex sting
   os << hex;
-  for (uint8_t byte: input)
-    os << static_cast<int>(byte);
+  for (uint8_t byte: input) {
+    int val = static_cast<int>(byte);
+    if (val<10)
+      os << 0;
```
axic: Please use something like this (from https://stackoverflow.com/questions/5760252/how-can-i-print-0x0a-instead-of-0xa-using-cout):
`setfill('0') << setw(2) << static_cast<int>(byte)`

PR #162, opened 2018-03-15
[WIP] Configure Hunter for Hera so it can be built as a standalone library

esteve:
This is just so Hera can have its own dependencies independently from `cpp-ethereum`, mostly to address https://github.com/ethereum/cpp-ethereum/pull/4891#discussion_r174432875


esteve:
@axic @chfast this PR will need some work so that it can both built as standalone and embedded into `cpp-ethereum`, but just wanted to gather feedback to know if it's worth pursuing this.

chfast:
Considering it is adding the same CMake code snippets, I'd go with using the Cable library.

Check out such modification in ethminer: https://github.com/ethereum-mining/ethminer/pull/808.

What do you think @axic ?

chfast:
I'm not sure Hunter is needed here at all. Does Hera have any dependencies other than binaryen?

axic:
It is only binaryen atm, but will be extended with wabt and wavm soon. I'd happy just using cable, if anything.

axic:
@chfast can we reuse this for building as a DLL?

axic:
Can this be closed then? Can you add such a CI task so we're sure it works (or I guess it will be part of your upcoming PR using a pre-built cpp-ethereum with Hera).

esteve:
Yeah, this can be closed, thanks for keeping track of this PR, I forgot about it.

PR #165, opened 2018-03-19
Zero-pad every byte in EVM string, not entire string

jwasinger:
Fix for EVM bytecode mangling issue that wasn't handled in https://github.com/ewasm/hera/pull/160


axic:
Merging this as it was passing tests before the rebase and the test suite seems to be broken now.

PR #167, opened 2018-03-26
Get rid of variable length array

chfast:


src/eei.cpp
```diff
@@ -426,7 +426,7 @@ string toHex(evm_uint256be const& value) {
       heraAssert(!(msg.flags & EVM_STATIC), "\"log\" attempted in static mode");
       heraAssert(numberOfTopics <= 4, "Too many topics specified");
 
-      evm_uint256be topics[numberOfTopics];
+      std::array<evm_uint256be, 4> topics{{}};
```
axic: We have the `std` namespace imported.

axic:
Why is `std::array` better?

chfast:
It's not about `std::array`. It's about `numberOfTopics` => `4`. `std::array` is just more convenient to use because it has `.begin()`, `.end()`, `.size()`, `.data()`...

PR #168, opened 2018-03-26
Use Cable

chfast:


axic:
What is the benefit of cable here?

chfast:
I thought we will need another CMake toolchain file for shared library, but it's not needed. So not big benefits, just the same compiler settings as in other C++ projects, and default build type being RelWithDebInfo.

esteve:
I'm not familiar with Cable, so I don't really have an informed opinion, but it seems like a look idea to match the same compiler settings as the rest of the C++ project.

axic:
@chfast merged in the `toHex` fix separately, please rebase.

axic:
Rebased.

@chfast is this PR still needed though?

axic:
@chfast ping

PR #170, opened 2018-03-27
Runtime debug options [WIP]

### LINKED ISSUE #169: Make debugging a runtime option via set_option

jakelang:
fixes #169

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -27,19 +27,8 @@
 
 using namespace std;
 using namespace wasm;
-
 #if HERA_DEBUGGING
-
-#define HERA_DEBUG cerr
-
-#else
-
-struct NullStream {
-  template<typename T> NullStream& operator<<(const T&) { return *this; }
-};
-
-#define HERA_DEBUG NullStream()
-
+using namespace HeraDebugging;
 #endif
 
 namespace HeraVM {
@@ -59,7 +48,9 @@ string toHex(evm_uint256be const& value) {
   void EthereumInterface::importGlobals(std::map<Name, Literal>& globals, Module& wasm) {
     (void)globals;
     (void)wasm;
-    HERA_DEBUG << "importGlobals\n";
+#if HERA_DEBUGGING
+    HeraDebug() << "importGlobals\n";
+#endif
   }
 
 #if HERA_DEBUGGING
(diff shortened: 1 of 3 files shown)


axic:
Sorry I wasn't clear (was just moving issues from the old testnet issue). I'd like to:
- keep the ifdef, but set the ifdef on by default
- have the runtime option, but set the debugging off by default

e.g. always compile support for debug, but don't run with debug. Also have the option to compile a much more optimised version (for benchmarking?)

src/eei.h
```diff
@@ -115,6 +119,9 @@ struct EthereumInterface : ShellExternalInterface {
   evm_message const& msg;
   std::vector<uint8_t> lastReturnData;
   ExecutionResult & result;
+  const bool debug;
+  /* FIXME */
+  // static std::ostream& DebugStream;
```
jakelang: @axic :
What I am doing here to replace HERA_DEBUG (previously defined by build options) is an ostream object reference that is defined for each instance of EthereumInterface, based on the debug flag. The issue I am having is that I want to use the NullStream struct we have defined already (the one without the large amount of useless inherited public methods), but this will not work because it is not an ostream like std::cerr. What is the best way to do this?

jakelang:
@axic oh, ok. I would still however like to hear the answer to the question I asked above about the debugstream ^^^

src/eei.h
axic: @chfast can you check this?

chfast: Can you give an example what is _not_ working?

jakelang: Nothing is broken at the moment, but I am not sure what is the best way to implement the HERA_DEBUG stream such that it works with runtime debug options.
What I do at the moment is give EthereumInterface an object reference to a stream, which is either NullStream or std::cerr depending on what value of "debug" is passed to the constructor. This doesn't work because NullStream is not an ostream and so DebugStream cannot refer to it.

The two solutions I have in mind are:
1. Implement NullStream with std::ostream as the base class and inherit a bunch of stuff we don't need
2. preprocessor hell

This is not super important because I am redoing this later, retaining compile-time options, but I would like to know how to approach this type of issue.

axic:
Why did you close it? :)

jakelang:
I am redoing this on a different branch, where i keep the compile-time options and don't do stupid things like pass the debug bool to each function that needs it.

axic:
You can force push into this branch.

jakelang:
@axic I think it makes sense to build this on top of the VM abstraction layer.
It would be much easier to do that first and then integrate debugging streams into that because it is much more strictly object-oriented than the mixed bag of imperative and OO style we have right now.

jakelang:
This will be implemented on top of the VM abstraction layer, when it is ready.

axic:
@jakelang plans to resume this or #239 ?

PR #171, opened 2018-03-28
Enable evm2wasm's evm trace output via CMake

jwasinger:
It occurs to me that we should be doing this at runtime.. Leaving this here anyways.


jwasinger:
Apologize for the terrible branch name XD

jwasinger:
ping @axic @chfast what do you think?

chfast:
There is no way to handle this with command line option? Where do you plan to use it?

jwasinger:
probably better as a CLI option,

PR #172, opened 2018-04-04
Migrate to EVMC

chfast:


src/eei.cpp
```diff
@@ -335,16 +335,16 @@ string toHex(evm_uint256be const& value) {
...
+      evmc_address address = loadUint160(addressOffset);
+      vector<uint8_t> codeBuffer(length);
+      size_t numCopied = context->fn_table->copy_code(context, &address, codeOffset, codeBuffer.data(), codeBuffer.size());
+      fill_n(&codeBuffer[numCopied], length - numCopied, 0);
 
       heraAssert(ffs(GasSchedule::copy) + (ffs(length) - 5) <= 64, "Gas charge overflow");
       heraAssert(numeric_limits<uint64_t>::max() - GasSchedule::extcode >= GasSchedule::copy * ((uint64_t(length) + 31) / 32), "Gas charge overflow");
       takeGas(GasSchedule::extcode + GasSchedule::copy * ((uint64_t(length) + 31) / 32));
-      // NOTE: code will be freed by the callee (client)
       // FIXME: optimise this so not vector needs to be created
```
axic: Comment is invalid now.

src/hera.cpp
```diff
@@ -370,13 +371,13 @@ void evm_destroy(evm_instance* instance)
...
+evmc_instance* hera_create()
 {
   hera_instance* instance = new hera_instance;
-  instance->destroy = evm_destroy;
-  instance->execute = evm_execute;
-  instance->set_option = evm_set_option;
-  return static_cast<evm_instance*>(instance);
+  instance->destroy = evmc_destroy;
+  instance->execute = evmc_execute;
+  instance->set_option = evmc_set_option;
```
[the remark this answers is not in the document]
chfast: I would name them without a prefix, ok?

src/eei.cpp
chfast: It is still valid. You can copy the code directly to the WASM memory.

axic: But then it should be moved to the point where the vector is created?

src/hera.cpp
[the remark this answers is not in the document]
axic: Well I'd prefix it with `instance_` then.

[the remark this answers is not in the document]
chfast: How about `hera_` prefix?

src/eei.cpp
```diff
@@ -335,16 +335,16 @@ string toHex(evm_uint256be const& value) {
...
-      size_t code_size = context->fn_table->get_code(&code, context, &address);
+      evmc_address address = loadUint160(addressOffset);
+      // FIXME: optimise this so not vector needs to be created
+      vector<uint8_t> codeBuffer(length);
+      size_t numCopied = context->fn_table->copy_code(context, &address, codeOffset, codeBuffer.data(), codeBuffer.size());
+      fill_n(&codeBuffer[numCopied], length - numCopied, 0);
 
       heraAssert(ffs(GasSchedule::copy) + (ffs(length) - 5) <= 64, "Gas charge overflow");
       heraAssert(numeric_limits<uint64_t>::max() - GasSchedule::extcode >= GasSchedule::copy * ((uint64_t(length) + 31) / 32), "Gas charge overflow");
       takeGas(GasSchedule::extcode + GasSchedule::copy * ((uint64_t(length) + 31) / 32));
```
axic: These assertions + gas charging should be before `copy_code`.

chfast: Moved.

axic:
@chfast any idea why tests are failing with this version?

chfast:
I assumed this is a failure in the main branch.

axic:
It is not failing on other PRs

chfast:
Depends on https://github.com/ewasm/hera/pull/184.

axic:
@chfast shouldn't this be updated to the latest EVMC version? I tried that yesteday but Hunter complained and cmake didn't even finish running.

chfast:
That would require more changes and even this change does not pass all the tests yet.

axic:
@ewasm/ewasm-team can you please remove the expectation of `revert` from any call in the test suite (for the moment)?

src/hera.cpp
```diff
@@ -268,21 +269,21 @@ void execute(
...
+evmc_result hera_execute(
+  evmc_instance *instance,
+  evmc_context *context,
+  enum evmc_revision rev,
+  const evmc_message *msg,
+  const uint8_t *code,
+  size_t code_size
+) {
+  evmc_result ret;
+  memset(&ret, 0, sizeof(evmc_result));
```
axic: Is this actually required in C++ or will `ret` be zero initialised anyway?

chfast: It's not 0 initialized (because it does not have a constructor, it's a C struct).
But in C++ you can do `evmc_result ret = {};` or `evmc_result ret{}`.
In C you can do `evmc_result ret = {EVM_FAILURE}`.

PR #173, opened 2018-04-10
Fix Circle CI build env

chfast:


axic:
What problem is this fixing?

PR #175, opened 2018-04-13
use mkstemp and print debug log for evm2wasmjs

cdetrio:


src/hera.cpp
```diff
@@ -138,12 +138,16 @@ string mktemp_string(string pattern) {
   const unsigned long len = pattern.size();
   char tmp[len];
   strncpy(tmp, pattern.data(), len);
-  if (!mktemp(tmp))
+  if (!mkstemp(tmp))
```
axic: `mkstemp` cannot be used here since it opens the file and it will never be closed, as a result it cannot be deleted, because one lingering open file descriptor will be tied to it. It will only be closed+deleted when cpp-ethereum is stopped.

What do you need to use `mkstemp` instead of `mktemp`?

cdetrio: CircleCI's servers do not like mktemp:
```
Calling evm2wasm.js (input 8 bytes)...
Calling evm2wasm.js with command: evm2wasm.js -e �tmp/hera.evm2wasm.evm.XXXXXX -o /tmp/hera.evm2wasm.wasm.VXkVHs
evm2wasm.js process.argv: [ '/usr/local/bin/node',
  '/home/builder/evm2wasm/bin/evm2wasm.js',
  '-e' ]
evm2wasm.js argv: { _: [], e: true }
evm2wasm.js inputFile: true
evm2wasm.js calling fs.readFileSync...
Error: TypeError: path must be a string or Buffer
InternalError: Transcompiling via evm2wasm failed
```

With this patch (using mkstemp rather than mktemp) on CircleCI servers:
```
Calling evm2wasm.js (input 8 bytes)...
Calling evm2wasm.js with command: evm2wasm.js -e /tmp/hera.evm2wasm.evm.XXXXXX -o /tmp/hera.evm2wasm.wasm.EAxt9V
evm2wasm.js process.argv: [ '/usr/local/bin/node',
  '/home/builder/evm2wasm/bin/evm2wasm.js',
  '-e',
  '/tmp/hera.evm2wasm.evm.XXXXXX',
  '-o',
  '/tmp/hera.evm2wasm.wasm.EAxt9V' ]
evm2wasm.js argv: { _: [],
  e: '/tmp/hera.evm2wasm.evm.XXXXXX',
  o: '/tmp/hera.evm2wasm.wasm.EAxt9V' }
evm2wasm.js inputFile: /tmp/hera.evm2wasm.evm.XXXXXX
evm2wasm.js calling fs.readFileSync...
evm2wasm.js converting bytecode to buffer..
evm2wasm.js calling convert..
evm2wasm.js done (output 869 bytes)
Executing...
```

axic: From the linux manpages:
> The mktemp() function always returns template. If a unique name was created, the last six bytes of template will have been modified in such a way that the resulting name is unique (i.e., does not exist already) If a unique name could not be created, template is made an empty string.

While on BSD:
> The mktemp() and mkdtemp() functions return a pointer to the template on success and NULL on failure.

I bet that `�` in front is a `\0`. 

One of the bugs here is that we only check for `NULL`, but it should also check for `tmp[0] == 0`.

The other problem is finding the reason why it fails to create a temp filename..

axic: Also the treason we cannot use `mkstemp` is that we do not get the filename, only the file descriptor (tried to find a way to extract the name off a descriptor, but couldn't), yet we need to pass the filename on the comandline.

cdetrio: I'll try that fix.

Issue #164, opened 2018-03-16
Ensure evmTrace is compliant with other clients


axic:
I think these are the remaining issues (may need to be fixed in evm2wasm directly):
- `pc` is off by one,
- need to figure out how to calculate the `gasLeft` correctly (I think #163 and https://github.com/ewasm/evm2wasm/issues/197 will fix it)
- need to display instruction names instead of opcodes

axic:
> `pc` is off by one,

Has been fixed by https://github.com/ewasm/evm2wasm/pull/200

> need to figure out how to calculate the `gasLeft` correctly

Has been fixed by https://github.com/ewasm/evm2wasm/pull/204 and the rest is tracked in #163.
> need to display instruction names instead of opcodes

Has been fixed by #179

PR #179, opened 2018-04-14
[WIP] Add EVM instructions names to be used by evmTrace

axic:
Part of #164.


axic:
@chfast can you please review? I haven't added all the instructions yet.

axic:
@jwasinger @cdetrio do you want to try this?

src/evm-instructions.h
```diff
@@ -0,0 +1,31 @@
...
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+#pragma once
+
+namespace HeraVM {
+
+const std::map<uint8_t, std::string> evmInstructionNames;
```
chfast: This map will be dynamically initialized also when never used.

src/eei.cpp
```diff
@@ -150,19 +151,22 @@ string toHex(evm_uint256be const& value) {
...
       uint32_t cost = static_cast<uint32_t>(arguments[2].geti32());
       int32_t sp = arguments[3].geti32();
 
       HERA_DEBUG << "evmTrace\n";
 
       static constexpr int stackItemSize = sizeof(evm_uint256be);
       heraAssert(sp <= (1024 * stackItemSize), "EVM stack pointer out of bounds.");
+      heraAssert(opcode >= 0x00 && opcode <= 0xff, "Invalid EVM instruction.");
+
+      string opName = evmInstructionNames.count(static_cast<uint8_t>(opcode)) ? evmInstructionNames.at(static_cast<uint8_t>(opcode)) : "UNKNOWN";
```
chfast: You search twice here.

axic: `.at` would throw an exception the opcode is not found.

src/evm-instructions.h
axic: Is that a lot memory?

src/eei.cpp
axic: What would the code look like? Is that any better / more concise?

src/evm-instructions.h
axic: Any suggestions?

src/eei.cpp
chfast:
```
auto it = evmInstructionNames.find(static_cast<uint8_t>(opcode));
string opName = (it != emvInstructionNames.end()) ? it->second : "UNKNOWN";
```

src/evm-instructions.h
chfast: Flat array.
But the map can stay for now. I will think about having this in EVMC because every VM implementations needs it.

chfast: See https://github.com/ethereum/cpp-ethereum/issues/4072.

axic: Is that an array of `char*` and for the missing ones filled with `NULL` or `"UNKOWN"`? If it goes into EVMC it has to be C.

src/eei.cpp
axic: I always found the iterators not so readable, but it does seem to do less lookups this way.

axic:
Should I hold back for EVMC to contain the instruction list?

chfast:
No, we should merge this with map and possibly using iterator for lookup.

axic:
@jwasinger you want to try this as part of tracing?

src/evm-instructions.h
```diff
@@ -26,6 +26,77 @@
 
 namespace HeraVM {
 
-const std::map<uint8_t, std::string> evmInstructionNames;
-
+const std::map<uint8_t, std::string> evmInstructionNames = {
```
axic: The actual values are in the `.cpp` file :)

jwasinger: If the map is declared const in the header, does that mean it can be reassigned in the cpp file?

When I use this PR with tracing, `evmInstructionNames.size()` is 0.

jwasinger: I moved the map into `evm-instructions.h` and removed `evm-instructions.cpp` as this seemed to be the only way to keep the map `const`.

axic: The problem with having it in a header it cannot be included in multiple files, unless marked `static` and in that case it is duplicated in the bytecode.

jwasinger: Ah okay.

jwasinger: @axic I got it working by adding `extern` to the header declaration.

jwasinger:
No, it needs a few more opcodes.

src/evm-instructions.cpp
```diff
@@ -0,0 +1,116 @@
...
+	{0xa4, "LOG4"},
+	{0xf0, "CREATE"},
+	{0xf1, "CALL"},
+	{0xf2, "CALLCODE"},
+	{0xf3, "RETURN"},
+	{0xf4, "DELEGATECALL"},
+	{0xf5, "CALLBLACKBOX"},
+	{0xfa, "STATICCALL"},
+	{0xfd, "REVERT"},
+	{0xff, "SUICIDE"},
```
axic: Should go with `SELFDESTRUCT`.

src/evm-instructions.cpp
```diff
@@ -0,0 +1,116 @@
...
+	{0xa1, "LOG1"},
+	{0xa2, "LOG2"},
+	{0xa3, "LOG3"},
+	{0xa4, "LOG4"},
+	{0xf0, "CREATE"},
+	{0xf1, "CALL"},
+	{0xf2, "CALLCODE"},
+	{0xf3, "RETURN"},
+	{0xf4, "DELEGATECALL"},
+	{0xf5, "CALLBLACKBOX"},
```
axic: This is a non-Byzantium opcode.

src/evm-instructions.cpp
axic: Also missing `INVALID`.

src/evm-instructions.cpp
```diff
@@ -0,0 +1,150 @@
...
+	{0x65, "PUSH6"},
+	{0x66, "PUSH7"},
+	{0x67, "PUSH8"},
+	{0x68, "PUSH9"},
+	{0x69, "PUSH10"},
+	{0x6a, "PUSH11"},
+	{0x6b, "PUSH12"},
+	{0x6c, "PUSH13"},
+	{0x6d, "PUSH14"},
+
```
axic: Where's `PUSH15`?

axic:
@jwasinger pushed my squashed commits. Can you have a look your commit and keep the style of the original .cpp file?

jwasinger:
sure

PR #183, opened 2018-04-16
Cleanup exceptions

chfast:


src/eei.cpp
```diff
@@ -866,7 +866,7 @@ string toHex(evm_uint256be const& value) {
     evm_uint256be balance;
     context->fn_table->get_balance(&balance, context, &msg.destination);
     if (safeLoadUint64(balance) < safeLoadUint64(value))
-      throw new OutOfGasException();
+      throw OutOfGasException();
```
axic: If `OutOfGasException{}` is the preferred style (see #184), then it should be changed in all instances in this PR.

axic:
If according to #184 it makes sense to making the inheritance of `std::exception` public, then it should be done in this PR for the rest of the exceptions.

src/hera.cpp
```diff
@@ -339,14 +339,14 @@ evm_result evm_execute(
 
     ret.status_code = result.isRevert ? EVM_REVERT : EVM_SUCCESS;
     ret.gas_left = result.gasLeft;
-  } catch (OutOfGasException) {
+  } catch (OutOfGasException const&) {
     ret.status_code = EVM_OUT_OF_GAS;
-  } catch (InternalErrorException &e) {
+  } catch (InternalErrorException const &e) {
```
axic: Also please move the ampersand next to const: `const& e`

src/eei.cpp
chfast: It's preferred in https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es23-prefer-the--initializer-syntax so I would use it.

chfast:
Suggestions applied.

axic:
Was this commit message "Do you throw a pointer to an exception object" meant to say "Do not throw.." ?

chfast:
Yes.

src/eei.cpp
```diff
@@ -866,7 +866,7 @@ string toHex(evm_uint256be const& value) {
     evm_uint256be balance;
     context->fn_table->get_balance(&balance, context, &msg.destination);
     if (safeLoadUint64(balance) < safeLoadUint64(value))
-      throw new OutOfGasException();
+      throw OutOfGasException{};
```
axic: @ewasm/testing should create a test case for this, added an issue: https://github.com/ewasm/tests/issues/22

PR #184, opened 2018-04-16
Correctly report Static Mode Violation

chfast:


src/eei.cpp
```diff
@@ -426,7 +426,9 @@ string toHex(evm_uint256be const& value) {
 
       HERA_DEBUG << "log " << hex << dataOffset << " " << length << " " << numberOfTopics << dec << "\n";
 
-      heraAssert(!(msg.flags & EVM_STATIC), "\"log\" attempted in static mode");
+      if (msg.flags & EVM_STATIC)
+        throw StaticModeViolation{};
```
axic: I'm not sure how beneficial it would be, but I think it may make sense keeping the function name which cause the violation.

chfast: I can add it to the exception, but there is no place where to use it (we should be moving from cout to proper tracing).

axic: We have those `HERA_DEBUGGING` lines. At least add `TODO` or `FIXME` where the exception is declared as a reminder. Plus create an issue for it.

src/eei.h
```diff
@@ -39,6 +39,11 @@ class OutOfGasException : public std::exception {
   const char* what() const noexcept override { return "Out of gas."; }
 };
 
+class StaticModeViolation : public std::exception {
+public:
+  const char* what() const noexcept override { return "Static mode violation"; }
```
axic: Dot missing (if want to be consistent with OOG).

chfast:
Review again.

src/hera.cpp
```diff
@@ -341,6 +341,8 @@ evm_result evm_execute(
     ret.gas_left = result.gasLeft;
   } catch (OutOfGasException const&) {
     ret.status_code = EVM_OUT_OF_GAS;
+  } catch (StaticModeViolation const&) {
+    ret.status_code = EVM_STATIC_MODE_ERROR;
```
axic: Actually, since you added the reason, it makes sense to put a `HERA_DEBUGGING` case here displaying the reason.

PR #187, opened 2018-04-16
 Move exceptions into its own file

axic:


src/exceptions.h
```diff
@@ -0,0 +1,51 @@
...
+
+class InternalErrorException : public std::exception {
+public:
+  explicit InternalErrorException(std::string _msg): msg(std::move(_msg)) {}
+  const char* what() const noexcept override { return msg.c_str(); }
+private:
+  std::string msg;
+};
+
+#define heraAssert(condition, msg) { \
```
axic: I'm also thinking about making this a function and having two overloads: this one and one with the exception type.

Or just introduce `ensureCondition(condition, exception)` for any non-internalError case?

PR #191, opened 2018-04-23
Ensure that balance/value is checked for as 128 bits and not 64 bits

### LINKED ISSUE #190: Value exceeds 64 bits

```
100%
Executing...
importGlobals
Executing...
importGlobals
```

axic:
Part of #190.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -872,15 +872,15 @@ string toHex(evmc_uint256be const& value) {
   {
     evmc_uint256be balance;
     context->fn_table->get_balance(&balance, context, &msg.destination);
-    if (safeLoadUint64(balance) < safeLoadUint64(value))
+    if (safeLoadUint128(balance) < safeLoadUint128(value))
       throw OutOfGasException{};
   }
 
-  uint64_t EthereumInterface::safeLoadUint64(evmc_uint256be const& value)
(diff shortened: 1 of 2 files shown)


axic:
@hugo-dc here's a fix if you want to try

hugo-dc:
Still getting an error:

InternalError: Value exceeds 128 bits.

It finds a non-zero byte in position 13 here: https://github.com/ewasm/hera/blob/3375c7eef8eef6ce42094c20ae12acf07710709a/src/eei.cpp#L902

axic:
As mentioned in the issue there is still an endianess problem.

hugo-dc:
LoadMemory loads 16 bytes of data into the first 16 bytes of `value`. When I send this value: `\20\a1\07\00\00\00\00\00\00\00\00\00\00\00\00\00`, I get the error I mentioned above (found a non-zero byte at position 13), if I send this value: `\00\00\00\00\00\00\00\00\00\00\00\00\00\07\a1\20` I get the same error, now indicating that the non-zero byte is in position 0.

```
i: 0
InternalError: Value exceeds 128 bits.
```

If I try to call sending 32 bytes it doesn't work either, because LoadMemory only reads 16 bytes (it would read only zeroes).

axic:
I'm merging this because it fixes one problem, 64 -> 128 bits. Lets fix endianess separately.

PR #192, opened 2018-04-23
Upgrade EVMC

chfast:


src/hera.cpp
```diff
@@ -54,7 +54,7 @@ struct hera_instance : evmc_instance {
   bool use_evm2wasm_js_trace = false;
 #endif
 
-  hera_instance() : evmc_instance({EVMC_ABI_VERSION, nullptr, nullptr, nullptr}) {}
+  hera_instance() : evmc_instance({EVMC_ABI_VERSION, "hera", "0.0.0", nullptr, nullptr, nullptr}) {}
```
axic: So actually cable would be useful here to have the buildinfo struct: )

axic:
Fails building:
```
/home/builder/cpp-ethereum/hera/src/eei.h:29:10: fatal error: 'evmc/evmc.h' file not found
#include <evmc/evmc.h>
         ^~~~~~~~~~~~~
```

jakelang:
should just be `<evmc.h>` since it is already a subdirectory according to cmake

chfast:
I just have to upgrade EVMC in cpp-ethereum as well to test Hera.

chfast:
cpp-ethereum upgraded as well. https://github.com/ethereum/cpp-ethereum/pull/4976
This is a circular dependency, but we should be fine.

@jakelang I moved the evmc header from `include/evmc.h` to `include/evmc/evmc.h`. In other projects having an include prefix like `#include <project-name/header.h` works better. Here we have single header so this is not crucial, but still I try to avoid having very different project files layout if possible.

chfast:
Tests passed.

PR #193, opened 2018-04-24
Fix CI cloning an already merged branch in cpp-ethereum

jakelang:
#4976 on cpp-ethereum has already been merged. Update CI here to use branch develop again.


jakelang:
replaced by #196

PR #194, opened 2018-04-25
Metering as a runtime option

### LINKED ISSUE #181: Move metering contract to a runtime option

jakelang:
Fixes #181 
needs ci to be fixed for tests to run #193

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -49,6 +49,7 @@ using namespace HeraVM;
 
 struct hera_instance : evmc_instance {
   bool fallback = false;
+  bool metering = false;
 #if HERA_EVM2WASM
   bool use_evm2wasm_js = false;
   bool use_evm2wasm_js_trace = false;
@@ -72,7 +73,6 @@ bool hasWasmPreamble(vector<uint8_t> const& _input) {
     _input[7] == 0;
 }
 
-#if HERA_METERING_CONTRACT || HERA_EVM2WASM
 vector<uint8_t> callSystemContract(
   evmc_context* context,
   evmc_address const& address,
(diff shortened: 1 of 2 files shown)


src/hera.cpp
```diff
@@ -53,6 +53,9 @@ struct hera_instance : evmc_instance {
   bool use_evm2wasm_js = false;
   bool use_evm2wasm_js_trace = false;
 #endif
+#if HERA_METERING_CONTRACT
```
axic: IIRC my idea in the issue was to remove the `ifdef`s entirely and just rely on this runtime option.

src/hera.cpp
```diff
@@ -390,6 +396,12 @@ int hera_set_option(
     hera->use_evm2wasm_js_trace = strcmp(value, "true") == 0;
     return 1;
   }
+#endif
+#if HERA_METERING_CONTRACT
+  if(strcmp(name, "metering-contract") == 0) {
```
axic: I'd just call this `metering`.

jakelang:
@axic the option is now documented in the readme.

src/hera.cpp
```diff
@@ -325,7 +320,8 @@ evmc_result hera_execute(
 
       if (msg->kind == EVMC_CREATE && !result.isRevert && hasWasmPreamble(result.returnValue)) {
```
axic: I'd just add the `hera->metering` check here to simplify the code (and avoid a copy).

axic: Perhaps its good to keep the `assert > 5`, so in that case must use `move` on `returnValue`.

axic:
Changed it myself.

PR #196, opened 2018-04-27
CI: Use cpp-ethereum with EVMC status code passthrough

chfast:
Closes #193.


axic:
@hugo-dc can you refill the tests again to support the revert status code? :)

axic:
Thank you @hugo-dc !

PR #198, opened 2018-05-09
make bounds checking be little-endian

jwasinger:


src/eei.cpp
```diff
@@ -903,7 +903,7 @@ string toHex(evmc_uint256be const& value) {
 
   bool EthereumInterface::exceedsUint128(evmc_uint256be const& value)
   {
-    for (unsigned i = 0; i < 16; i++) {
+    for (unsigned i = 16; i < 32; i++) {
```
axic: This will actually make it big endian.

src/eei.cpp
```diff
@@ -894,7 +894,7 @@ string toHex(evmc_uint256be const& value) {
 
   bool EthereumInterface::exceedsUint64(evmc_uint256be const& value)
   {
-    for (unsigned i = 0; i < 24; i++) {
+    for (unsigned i = 24; i < 32; i++) {
```
axic: This will actually make it big endian.

axic:
Closing this as I think we agreed this isn't the right fix.

PR #201, opened 2018-05-10
Support optional message in OutOfGasException

axic:


src/exceptions.h
```diff
@@ -28,7 +28,12 @@ namespace HeraVM {
 
 class OutOfGasException : public std::exception {
 public:
-  const char* what() const noexcept override { return "Out of gas."; }
+  explicit OutOfGasException(std::string const& _msg = "Out of gas."):
+    msg(std::move(_msg))
```
chfast: Nope. You cannot move a `const` object (this still will call the copy constructor not the move constructor).
https://www.youtube.com/watch?v=ZKaoR3dP9uM

The canonical way would be:
```
explicit OutOfGasException(std::string _msg = "Out of gas."):
    msg(std::move(_msg))
```

src/exceptions.h
```diff
@@ -28,7 +28,12 @@ namespace HeraVM {
...
+  {}
+  const char* what() const noexcept override { return msg.c_str(); }
+private:
+  std::string msg;
```
chfast: If you don't plan to have runtime created messages, you can just keep the reference to a string literal.

src/exceptions.h
axic: It's a typo, there were multiple iterations and added the move at the last minute.

src/exceptions.h
axic: See the next PR depending on this for the reason.

src/exceptions.h
axic: Fixed.

PR #202, opened 2018-05-10
Add ensure condition

### LINKED ISSUE #188: Cleanup exception handling
See https://github.com/ewasm/hera/pull/187#discussion_r181896684

Probably makes sense keeping `heraAssert` for internal exceptions and introducing a new function `ensureCondition(condition, exception)` to throw specific exceptions, such as static mode violation, out of bounds memory access, etc.

axic:
Depends on #201. Closes #188.

This is the first in introducing more granular exception and a nice way to have assertions for them.


axic:
Please note this PR is not finished yet (hence the WIP marker).

PR #206, opened 2018-05-13
Only reject contract code without wasm preamble if evm2wasm is not enabled

jwasinger:


axic:
Next time can you please follow the convention of capitalised commit messages? :)

PR #208, opened 2018-05-15
Introduce InvalidMemoryAccess exception and use where relevant

axic:


src/eei.cpp
```diff
@@ -777,7 +777,7 @@ string toHex(evmc_uint256be const& value) {
 
   void EthereumInterface::loadMemory(uint32_t srcOffset, uint8_t *dst, size_t length)
   {
-    heraAssert((srcOffset + length) > srcOffset, "Out of bounds (source) memory copy.");
+    ensureCondition((srcOffset + length) >= srcOffset, InvalidMemoryAccess, "Out of bounds (source) memory copy.");
```
axic: I think this was should be `>=` as the others, to support `length = 0`.

PR #210, opened 2018-05-15
Introduce ContractValidationFailure exception and use where relevant

axic:


src/exceptions.h
```diff
@@ -36,6 +36,16 @@ class OutOfGasException : public std::exception {
   std::string msg;
 };
 
+class ContractValidationFailure : public std::exception {
```
axic: Someone should refactor this file, separately 😉

Issue #116, opened 2018-01-24
Validate imports prior to execution


axic:
Follow up of #103.

Only known imports with the correct number of arguments (perhaps also the types of arguments) should be accepted.

PR #211, opened 2018-05-15
Assert argument counts in the EEI

axic:
Part of #116.


axic:
@hugo-dc it seems some of the imports use an incompatible number of arguments. Perhaps `storageStore` ?

PR #212, opened 2018-05-15
Validate basic ECI rules

### LINKED ISSUE #116: Validate imports prior to execution

axic:
Fixes #116.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -240,6 +240,40 @@ vector<uint8_t> evm2wasm(evmc_context* context, vector<uint8_t> const& input) {
   return ret;
 }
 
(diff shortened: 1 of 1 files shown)


axic:
#211 should only have assertions which are runtime failures, but this PR should properly validate imports and return a `ContractValidationFailure` if there's a mismatch.

axic:
@hugo-dc do you have any idea why the ewasm tests would fail after validation? Are they exporting other things? Are they not exporting memory in every case?

jwasinger:
We should have a test case to catch this.

PR #215, opened 2018-05-15
 Return EVM_INTERNAL_ERROR for InternalErrorException

### LINKED ISSUE #185: Move all OOG related cases to throw OOG and not a generic assertion

axic:
Fixes #185.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -934,7 +934,8 @@ string toHex(evmc_uint256be const& value) {
 
   void EthereumInterface::storeUint128(evmc_uint256be const& src, uint32_t dstOffset)
   {
-    heraAssert(!exceedsUint128(src), "Value at src cannot exceed 2^128-1");
+    // TODO: use a specific error code here?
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -934,7 +934,8 @@ string toHex(evmc_uint256be const& value) {
 
   void EthereumInterface::storeUint128(evmc_uint256be const& src, uint32_t dstOffset)
   {
-    heraAssert(!exceedsUint128(src), "Value at src cannot exceed 2^128-1");
+    // TODO: use a specific error code here?
+    ensureCondition(!exceedsUint128(src), OutOfGasException, "Value exceeds 128 bits.");
```
chfast: I assume this ends with `EVMC_OUT_OF_GAS`. It would be better to return `EVMC_FAILURE` to indicate it's some kind of unspecified failure.

axic: Correct, there's a todo mentioned there. I want to postpone this one after the exception refactoring.

PR #217, opened 2018-05-16
[WIP]  Do not keep copy of local code in EthereumInterface

axic:
This fixes `getCodeSize` and `codeCopy` for evm2wasm.


src/eei.cpp
```diff
@@ -355,7 +355,14 @@ string toHex(evmc_uint256be const& value) {
         "Gas charge overflow"
       );
       takeGas(GasSchedule::verylow + GasSchedule::copy * ((uint64_t(length) + 31) / 32));
-      storeMemory(code, codeOffset, resultOffset, length);
+
+      evmc_address address = msg.destination;
+      // FIXME: optimise this so not vector needs to be created
```
jwasinger: @axic Can this PR be merged as is (with an issue noting this optimization)?

PR #219, opened 2018-05-17
Add proper gas stipend and 63/64th rules for CALL*

cdetrio:
building on https://github.com/ewasm/hera/pull/217 for the test case `stCallCodes/callcall_00` with evm2wasm.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -664,7 +664,6 @@ string toHex(evmc_uint256be const& value) {
       call_message.destination = loadUint160(addressOffset);
       call_message.flags = msg.flags;
       call_message.code_hash = {};
-      call_message.gas = gas - (gas / 64);
       call_message.depth = msg.depth + 1;
 
       if (import->base == Name("call") || import->base == Name("callCode")) {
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -718,12 +730,32 @@ string toHex(evmc_uint256be const& value) {
...
+        //takeGas(GasSchedule::callNewAccount);
+      }
+      if (!isZeroUint256(call_message.value)) {
+        extra_gas += GasSchedule::valuetransfer;
+        //takeGas(GasSchedule::valuetransfer);
+      }
+      extra_gas += GasSchedule::call;
+      //takeGas(GasSchedule::call);
+
+      int64_t gas_available = result.gasLeft - extra_gas;
```
jwasinger: The `63/64` rule should be applied after the call is made.

jwasinger: Nevermind... it is applied before.

src/eei.cpp
```diff
@@ -740,11 +776,11 @@ string toHex(evmc_uint256be const& value) {
 
       switch (call_result.status_code) {
       case EVMC_SUCCESS:
-        return Literal(uint32_t(0));
+        return Literal(uint32_t(1));
```
axic: The spec currently says 0 for success, 1 for failure and 2 for revert (since it is a POSIX style enum and not a boolean anymore).

cdetrio: ok, it makes sense to conver this in evm2wasm.  since its EVM that uses 1 for success and 0 for failure (opposite of the POSIX standard)

chfast: I believe you can do `uint32_t(EVMC_FAILURE)`.

axic: How is it related to `EVMC` error codes?

jwasinger:
Is this PR likely to get approved?  Or do we we not want to include these call gas calculation rules in the EEI spec (i.e. move them to evm2wasm)?

cdetrio:
https://eips.ethereum.org/EIPS/eip-150

> Define “all but one 64th” of N as N - floor(N / 64).
If a call asks for more gas than the maximum allowed amount (i.e. the total amount of gas remaining in the parent after subtracting the gas cost of the call and memory expansion), do not return an OOG error; instead, if a call asks for more gas than all but one 64th of the maximum allowed amount, call with all but one 64th of the maximum allowed amount of gas (this is equivalent to a version of EIP-901 plus EIP-1142). CREATE only provides all but one 64th of the parent gas to the child call.

axic:
@hugo-dc do you want to update tests to follow EIP150 gas calulcations?

cdetrio:
only a cursory glance but looks much leaner.  i'd like to add the test cases for stCallCodes/callcall_00 and stCallCodes/callcallcallcode_01 to evm2wasm-tests.sh (on circleCI) to this PR before merging

cdetrio:
appears to pass stCallCodes/callcall_00 and callcallcodes_01 with evm2wasm, so LGTM

Issue #221, opened 2018-05-18
Memory copying should write zeros for bytes that are out of bounds in the source location


jwasinger:
See this trace of randomStatetest283 from evm2wasm:
https://gist.github.com/jwasinger/f85a735066491d56f52513a1cffddeeb

jwasinger:
https://github.com/ewasm/hera/pull/222

lrettig:
Looks like this is no longer an issue per https://github.com/ewasm/hera/pull/222#issuecomment-390824463, also https://github.com/ewasm/tests/issues/34 and https://github.com/ewasm/evm2wasm/issues/250. Can we close this?

PR #222, opened 2018-05-21
Memory load/store:  bytes that are out of bounds should be written/read as 0

jwasinger:
**WIP**

Fixes the test case `randomStatetest283`.

Fixes https://github.com/ewasm/hera/issues/221


axic:
I think we tried to break away from that EVM nonsense in ewasm :)

axic:
I think practically this can be implemented in evm2wasm with `returndatasize` + zero-filling the different between the expected length of the call return area.

jwasinger:
@axic Okay, closing this PR and we can continue tracking this in: https://github.com/ewasm/evm2wasm/issues/250

PR #228, opened 2018-05-26
don't use mktemp for hera evm2wasm.js

jwasinger:
Broken on debian-based environments.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -133,6 +133,7 @@ vector<uint8_t> sentinel(evmc_context* context, vector<uint8_t> const& input)
 }
 
 // NOTE: assumes that pattern doesn't contain any formatting characters (e.g. %)
+/*
 string mktemp_string(string pattern) {
   const unsigned long len = pattern.size();
   char tmp[len + 1];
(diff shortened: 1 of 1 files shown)


axic:
@jwasinger why is this need?

jwasinger:
Try evm2wasm.js on a debian-environment.  the cpp-build-env image is debian (or maybe alpine?)

axic:
You need to put in some kind of log showing what and how fails.

jwasinger:
```
#include <stdio.h>
#include <string>
#include <string.h>
#include <iostream>
#include <stdlib.h>

using namespace std;

string mktemp_string(string pattern) {
  const unsigned long len = pattern.size();
  char tmp[len];
  strncpy(tmp, pattern.data(), len);
  if (!mktemp(tmp))
     return string();
  return string(tmp, len);
}

int main(int argc, char **argv) {
  string s = mktemp_string("/tmp/hera.evm2wasm.evm.XXXXXX");
  string s2 = mktemp_string("/tmp/hera.evm2wasm.evm.XXXXXX");
  cout << s << endl << s2;
}
```

This doesn't work in affected environments.

axic:
This was fixed in #176.

PR #220, opened 2018-05-18
revert internal failure to EVMC_FAILURE.

jwasinger:
@cdetrio and I found that the evm2wasm stack tests were failing because of this line.  This reverts the part of the change in https://github.com/ewasm/hera/pull/215 to fix the issue.

PR #229, opened 2018-05-26
Consider wasm traps as non-internal errors

axic:
Closes #220.


jwasinger:
This doesn't appear to be fully working.  For example, in `stBadOpcode` tests all are failing because an internal EVM exception occurs in contract creation and the account is still created.

PR #232, opened 2018-05-27
Test evm2wasm in CircleCI (and add test script)

cdetrio:


scripts/evm2wasm-tests.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "running tests.sh inside working dir: $WORKING_DIR"
+
+echo "listing files:"
+ls -al
+
+echo "fetch ewasm tests."
+git clone https://github.com/ethereum/tests
+
+echo "fetch evm2wasm"
+git clone https://github.com/ewasm/evm2wasm.git
```
axic: Should use a fixed version of evm2wasm.

axic:
@cdetrio rebased this and it fails because the `evm2wasm` directory already exists. Can you change the script to use a different one?

cdetrio:
trying to just use the evm2wasm submodule directory.

PR #233, opened 2018-05-27
Support invoking the C++ evm2wasm directly

axic:


src/hera.cpp
```diff
@@ -50,6 +52,8 @@ using namespace HeraVM;
 enum hera_evm_mode {
   EVM_FALLBACK,
   EVM2WASM,
```
jakelang: This option is not needed anymore?

jakelang: Maybe rename this to EVM2WASM_CONTRACT for clarity

axic: Do you want to create a PR for that?

jakelang: @axic #235

PR #236, opened 2018-05-28
Properly handle rejecting EVM bytecode

axic:
Fixes #213.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -50,6 +50,7 @@ using namespace wasm;
 using namespace HeraVM;
 
 enum hera_evm_mode {
+  EVM_REJECT,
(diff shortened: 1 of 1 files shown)


src/hera.cpp
```diff
@@ -357,9 +358,11 @@ evmc_result hera_execute(
       case EVM_FALLBACK:
         ret.status_code = EVMC_REJECTED;
         return ret;
-      default:
+      case EVM_REJECT:
         ret.status_code = EVMC_FAILURE;
         return ret;
+      default:
+        heraAssert(false, "");
```
jakelang: Maybe make the message "Invalid execution mode" or something, even if this will probably never happen

PR #239, opened 2018-05-28
Runtime option for debugging messages

jakelang:
Implement a "stream" for debug messages that just pipes the input to the right stream based on a boolean member.

Depends on and merges into #148


jakelang:
will do this later

PR #240, opened 2018-05-29
Improve storagestore debug output

lrettig:
Print the storage key and value being stored as well as the memory addresses

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -562,13 +562,15 @@ inline int64_t maxCallGas(int64_t gas) {
       uint32_t pathOffset = arguments[0].geti32();
       uint32_t valueOffset = arguments[1].geti32();
 
-      HERA_DEBUG << "storageStore " << hex << pathOffset << " " << valueOffset << dec << "\n";
+      HERA_DEBUG << "storageStore offsets path " << hex << pathOffset << " value " << valueOffset << dec << "\n";
 
       ensureCondition(!(msg.flags & EVMC_STATIC), StaticModeViolation, "storageStore");
 
       evmc_uint256be path = loadUint256(pathOffset);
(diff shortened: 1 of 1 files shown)


jakelang:
@axic can you give @lrettig write access to this repo? Working with forks is insufferable

src/eei.cpp
```diff
@@ -562,14 +562,16 @@ inline int64_t maxCallGas(int64_t gas) {
...
-      HERA_DEBUG << "storageStore " << hex << pathOffset << " " << valueOffset << dec << "\n";
+      HERA_DEBUG << "storageStore offsets path " << hex << pathOffset << " value " << valueOffset << dec << "\n";
 
       ensureCondition(!(msg.flags & EVMC_STATIC), StaticModeViolation, "storageStore");
 
       evmc_uint256be path = loadUint256(pathOffset);
       evmc_uint256be value = loadUint256(valueOffset);
       evmc_uint256be current;
 
+      HERA_DEBUG << "storageStore contents path " << hex << toHex(path) << " value " << toHex(value) << dec << "\n";
```
jakelang: `toHex` is only defined when built with debugging, causing build to fail in release mode.

lrettig: What about https://github.com/lrettig/hera/blob/6a88a35fffce35f0fce389c042100ace9a97e3c3/src/eei.cpp#L194?

jakelang: @lrettig that's in the debug module which is only compiled in debug mode

lrettig: I find this insufferable. Lol. Either toHex should be moved out of the #ifdef, or else all `HERA_DEBUG` statements should be compiled out in prod mode. Anyway this isn't critical so I'm gonna close this for now.

PR #245, opened 2018-05-29
Build intx bignum library in CMake

### LINKED ISSUE #190: Value exceeds 64 bits

jakelang:
Fixes #190

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -24,6 +24,9 @@
 
 #include <stdexcept>
 #include <array>
+
+#include <intx.hpp>
+
 #include "eei.h"
 #include "exceptions.h"
 #include "evm-instructions.h"
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -977,11 +977,11 @@ inline int64_t maxCallGas(int64_t gas) {
     ensureCondition(safeLoadUint128(balance) >= safeLoadUint128(value), OutOfGasException, "Out of gas.");
   }
 
-  uint64_t EthereumInterface::safeLoadUint128(evmc_uint256be const& value)
+  intx::uint128 EthereumInterface::safeLoadUint128(evmc_uint256be const& value)
   {
     // TODO: use a specific error code here?
     ensureCondition(!exceedsUint128(value), OutOfGasException, "Value exceeds 128 bits.");
-    uint64_t ret = 0;
+    intx::uint128 ret = 0;
     for (unsigned i = 16; i < 32; i++) {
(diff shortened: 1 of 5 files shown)


axic:
@chfast can you review this?

jakelang:
@chfast `INTX_TESTING` doesn't exist on latest master branch yet.

jakelang:
@chfast just opened a PR on intx to make said option.

jakelang:
`badab6e` seems to be causing ninja to break in Debug mode with this error message: 
`ninja: error: 'intx::intx-NOTFOUND', needed by 'src/libhera.so', missing and no known rule to make it`
 I tried building with RelWithDebInfo mode as well as None which sticks to existing compiler options but these cause the compiler to fail with 
`src/CMakeFiles/hera.dir/build.make:147: *** target pattern contains no '%'.  Stop.`

chfast:
In Ninja you have to specify "byproducts" (as you did). Maybe some path there is incorrect. I would stick to Make because it's easier to start with.

For Make error, you have to check out what's in hera.dir/build.make:147.

src/eei.cpp
```diff
@@ -974,11 +977,11 @@ inline int64_t maxCallGas(int64_t gas) {
     ensureCondition(safeLoadUint128(balance) >= safeLoadUint128(value), OutOfGasException, "Out of gas.");
   }
 
-  uint64_t EthereumInterface::safeLoadUint128(evmc_uint256be const& value)
+  intx::uint128 EthereumInterface::safeLoadUint128(evmc_uint256be const& value)
```
chfast: You know that GCC / clang have `unsigned __int128`?

axic: @chfast would have been cool if you could have mentioned that when the original issue was created :)

But according to https://gcc.gnu.org/onlinedocs/gcc/_005f_005fint128.html :
> As an extension the integer scalar type __int128 is supported for targets which have an integer mode wide enough to hold 128 bits. 
> There is no support in GCC for expressing an integer constant of type __int128 for targets with long long integer less than 128 bits wide.

jakelang: So we can assume it works for x86-64 but not much else until we investigate

chfast: I wasn't aware that 128-bit is the max precision you need.

Anyway, the type is available on 64-bit targets. Because you don't support MSVC at the moment it can be used as a temporary solution.

PR #247, opened 2018-06-01
Use int128 compiler extension for ensureSenderBalance

### LINKED ISSUE #190: Value exceeds 64 bits

jakelang:
portability of this fix requires investigation

Closes #245. Fixes #190.


jakelang:
@ewasm/testing If anybody here has time to write a test to ensure that this fix works, please do

jwasinger:
@jakelang if you made the fix, why don't you write the test as well.

PR #249, opened 2018-06-02
Revert/return/selfdestruct to properly abort

axic:
Depends on https://github.com/ewasm/design/issues/72 (and https://github.com/ewasm/design/pull/87)

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -613,7 +613,7 @@ inline int64_t maxCallGas(int64_t gas) {
 
       result.isRevert = import->base == Name("revert");
 
-      return Literal();
+      throw EndExecution{};
     }
 
     if (import->base == Name("getReturnDataSize")) {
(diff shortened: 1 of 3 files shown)


axic:
@chfast can you review the implementation detail?

src/eei.cpp
```diff
@@ -619,7 +619,7 @@ inline int64_t maxCallGas(int64_t gas) {
 
       result.isRevert = import->base == Name("revert");
 
-      return Literal();
+      throw new EndExecution{};
```
chfast: Should be `throw EndExecution{}`. You want to throw an exception of type `EndExecution` not `EndExecution*`.

src/hera.cpp
```diff
@@ -326,7 +326,13 @@ void execute(
 
   Name main = Name("main");
   LiteralList args;
-  instance.callExport(main, args);
+
+  try {
+    instance.callExport(main, args);
+  } catch (EndExecution const&) {
```
chfast: This will not work because the exception type is `EndExecution*` (no match). Must be fixed in other place.

axic:
> I'm not a fan of using exceptions for control flow. I don't see the full picture, but I would avoid using C++ exceptions for any eWASM / EVM specified failures.

I don't think there is any other way in binaryen. This is not a failure, but aborting on `return` / `revert`, just like how `exit` works in wasm.

src/hera.cpp
axic: Fixed both.

axic:
The design change was merged and no test changes are required.

PR #250, opened 2018-06-02
Fix gas related to CREATE

lrettig:
Requires https://github.com/ewasm/tests/pull/44

Take base gas before doing 63/64 calc and refund remaining gas after call.
Also includes a bunch more gas-related debug output and a new debug flag to keep evm2wasm output files around for debugging.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -814,6 +814,9 @@ inline int64_t maxCallGas(int64_t gas) {
 
       context->fn_table->call(&create_result, context, &create_message);
 
+      /* Return unspent gas */
+      result.gasLeft += create_result.gas_left;
+
       if (create_result.status_code == EVMC_SUCCESS) {
         storeUint160(create_result.create_address, resultOffset);
         lastReturnData.clear();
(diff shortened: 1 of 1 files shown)


lrettig:
ewasm/tests#44 was just merged, we should rerun these tests. I get an error saying I can't re-trigger the tests since I don't have write permission. @axic could you please rerun? 🙏

axic:
I don't think it should have all this debugging as the output will go crazy. I'll remove them.

src/eei.cpp
```diff
@@ -790,6 +791,8 @@ inline int64_t maxCallGas(int64_t gas) {
 
       // NOTE: this must be declared outside the condition to ensure the memory doesn't go out of scope
       vector<uint8_t> contract_code;
+      HERA_DEBUG << "Taking GasSchedule::create" << endl;
+      takeInterfaceGas(GasSchedule::create);
```
axic: I don't think this is the correct place. Should be grouped with the other `takeInterfaceGas` call. Moved.

axic:
@lrettig have a look if you still ok with version.

PR #252, opened 2018-06-11
EVM instruction names

chfast:


src/eei.cpp
```diff
@@ -177,8 +178,9 @@ inline int64_t maxCallGas(int64_t gas) {
       heraAssert(sp <= (1024 * stackItemSize), "EVM stack pointer out of bounds.");
       heraAssert(opcode >= 0x00 && opcode <= 0xff, "Invalid EVM instruction.");
 
-      auto it = evmInstructionNames.find(static_cast<uint8_t>(opcode));
-      string opName = (it != evmInstructionNames.end()) ? it->second : "UNKNOWN";
+      const char* opName = evmc_get_instruction_name_table()[static_cast<uint8_t>(opcode)];
+      if (opName == nullptr)
+        opName = "UNDEFINED";
```
axic: Is `undefined` used by the tracing format? The code above used `unknown` previously.

chfast: I'm not sure tests depends on this, but the "undefined" name comes from https://github.com/ethereum/evmc/blob/master/include/evmc/evmc.h#L229.

src/eei.cpp
```diff
@@ -177,8 +178,9 @@ inline int64_t maxCallGas(int64_t gas) {
       heraAssert(sp <= (1024 * stackItemSize), "EVM stack pointer out of bounds.");
       heraAssert(opcode >= 0x00 && opcode <= 0xff, "Invalid EVM instruction.");
 
-      auto it = evmInstructionNames.find(static_cast<uint8_t>(opcode));
-      string opName = (it != evmInstructionNames.end()) ? it->second : "UNKNOWN";
+      const char* opName = evmc_get_instruction_name_table()[static_cast<uint8_t>(opcode)];
```
axic: I'd like to have a check that this only gets `byzantium` opcodes since that is what evm2wasm "supports". The only way to do that with the current API is to build a local array while calling `get_instruction_metrics` every time. See my comments here: https://github.com/ethereum/evmc/pull/33

axic:
Build and tests are failing.

jakelang:
@axic this is probably failing because it needs to be rebased for the create gas changes in #250

axic:
This will need to be changed after https://github.com/ethereum/evmc/pull/34 is merged. I think this one here is pending till then.

jakelang:
@chfast I think that the tests are failing build because cpp-ethereum has an outdated evmc submodule pointing to a commit without the `evmc::evmc` alias

chfast:
Updated cpp-ethereum, ewasm-tests is fine now.
Is failure of linux-clang-shared-release expected?

chfast:
The failure in `linux-clang-shared-release` happens in all other PRs. Please review!

axic:
@chfast is this now updated to the final evmc version for instruction names?

axic:
> The failure in `linux-clang-shared-release` happens in all other PRs. Please review!

I think someone broke the test suite again and left in debugging statements.

cc @hugo-dc @lrettig @jwasinger ?

src/eei.cpp
```diff
@@ -177,8 +178,10 @@ inline int64_t maxCallGas(int64_t gas) {
       heraAssert(sp <= (1024 * stackItemSize), "EVM stack pointer out of bounds.");
       heraAssert(opcode >= 0x00 && opcode <= 0xff, "Invalid EVM instruction.");
 
-      auto it = evmInstructionNames.find(static_cast<uint8_t>(opcode));
-      string opName = (it != evmInstructionNames.end()) ? it->second : "UNKNOWN";
+      const char* const* const opNamesTable = evmc_get_instruction_names_table(EVMC_BYZANTIUM);
```
axic: OMG, that looks hideous

hugo-dc:
@axic it is not passing one of the newer test cases, I'll take a look

src/eei.cpp
chfast: True. If you want I can change it to `const char* const*` :D

chfast:
Can I merge it now?

hugo-dc:
One of the new test case is failing `createNonzero`, I think we should remove it by now and investigate why it is not passing. cc @lrettig

lrettig:
@hugo-dc just ran `createNonzero` against this hera branch and it's passing for me. Can someone kick off these tests again? They were last run four days ago.

chfast:
Green!

src/eei.cpp
chfast: No, because you cannot return an array in C. It must be a const pointer to an item (which is `const char*`).

chfast: Here you can just use `auto` or `auto*`.

axic: Does that ensure constness though?

chfast: Yes. The best would be `const auto`.

PR #254, opened 2018-06-11
Refactor exceptions to share more code

axic:


axic:
@chfast would you mind looking into the refactoring and giving hints how to fix it? :)

jakelang:
@axic you must specify inherited constructors in the child class.
`using HeraException::HeraException` should do the trick

jakelang:
fixed it @axic

src/exceptions.h
```diff
@@ -28,16 +28,24 @@ namespace HeraVM {
 
 class HeraException : public std::exception {
 public:
-  explicit HeraException(std::string _msg): msg(std::move(_msg)) {}
```
axic: Does this needs to be removed?

axic: @jakelang @chfast ping ?

PR #255, opened 2018-06-17
Fix gas semantics for CALL and CALLCODE

lrettig:
If there is insufficient gas for the value being sent with the CALL,
rather than throwing an OOG exception, the operation should just fail.

Fixes an issue detected in test GeneralStateTests/stCallCreateCallCodeTest/callWithHighValue

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -727,15 +727,16 @@ inline int64_t maxCallGas(int64_t gas) {
           extra_gas += GasSchedule::callNewAccount;
       }
 
-      // this check is in EIP150 but not in the YellowPaper
+      // This check is in EIP150 but not in the YellowPaper
       takeInterfaceGas(extra_gas);
 
-      // retain one 64th gas (EIP150)
+      // This is the gas we are forwarding to the callee.
+      // Retain one 64th of it as per EIP150
       gas = std::min(gas, maxCallGas(result.gasLeft));
 
       takeInterfaceGas(gas);
 
-      // add gas stipend for value transfers
+      // Add gas stipend for value transfers
       if (!isZeroUint256(call_message.value))
         gas += GasSchedule::valueStipend;
 
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -717,13 +717,17 @@ inline int64_t maxCallGas(int64_t gas) {
       }
 
       evmc_result call_result;
-      int64_t extra_gas = 0;
 
-      if (import->base == Name("call") && !context->fn_table->account_exists(context, &call_message.destination))
-        extra_gas += GasSchedule::callNewAccount;
-      if (!isZeroUint256(call_message.value))
+      // Start with base call gas
+      int64_t extra_gas = GasSchedule::call;
+
+      // Charge valuetransfer gas if value is being transferred.
+      // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
+      if (!isZeroUint256(call_message.value)) {
         extra_gas += GasSchedule::valuetransfer;
-      extra_gas += GasSchedule::call;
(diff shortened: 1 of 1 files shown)


lrettig:
I don't know why the test is failing. I just checked and `createNonzero` is passing for me, locally. Also that test does not call `CALL` and these changes clearly only affect `CALL`. Weird.

jakelang:
@lrettig the same problem occurs on every PR. Looking into it right now

src/eei.cpp
```diff
@@ -744,6 +742,17 @@ inline int64_t maxCallGas(int64_t gas) {
 
       call_message.gas = gas;
 
+      if (import->base == Name("call") || import->base == Name("callCode")) {
+        try {
+          ensureSenderBalance(call_message.value);
+        } catch (OutOfGas const& e) {
```
axic: `e` is not used after this, shouldn't be named to avoid `unused variable` errors

src/eei.cpp
```diff
@@ -744,6 +742,17 @@ inline int64_t maxCallGas(int64_t gas) {
 
       call_message.gas = gas;
 
+      if (import->base == Name("call") || import->base == Name("callCode")) {
```
axic: Does this needs to be moved from the original position?

lrettig: Not sure what you mean by "moved." The original still exists at line 673. This line was added because this change only applies to `call` and `callCode`, not to the other host functions handled in this function.

axic: To me it seems the original line 686 was removed and was moved over here?

lrettig: I see what you mean. I misread which line you were referring to. Fair point. Let me move and test.

lrettig: I tested with the check in the original location. The test fails because no gas is consumed. We need to charge some gas in this case. See e.g. the logic at https://github.com/ethereum/py-evm/blob/master/evm/vm/logic/call.py#L65.

src/eei.cpp
```diff
@@ -744,6 +742,18 @@ inline int64_t maxCallGas(int64_t gas) {
...
+        try {
+          ensureSenderBalance(call_message.value);
+        } catch (OutOfGas) {
```
axic: This should be a const reference to avoid cloning exception objects, e.g. `OutOfGas const&`.

src/eei.cpp
axic: I'm not sure all this random logic matches with the YP, so frustrating.

src/eei.cpp
lrettig: Fixed

lrettig:
I think this is ready to merge, am I missing anything?

src/eei.cpp
```diff
@@ -733,12 +735,24 @@ inline int64_t maxCallGas(int64_t gas) {
 
       takeInterfaceGas(gas);
 
-      // add 2300 gas stipend for value transfers
+      // add gas stipend for value transfers
       if (!isZeroUint256(call_message.value))
-        gas += 2300;
+        gas += GasSchedule::valueStipend;
```
axic: Can you create a separate PR for this? It is just moving it into a variable.

axic: Or at least a separate commit, but a PR is fine, we can merge it quickly.

lrettig:
@axic thanks for the feedback on this one. I cleaned up the commit log, please take another look and let me know if it makes more sense now.

src/eei.cpp
```diff
@@ -743,6 +741,8 @@ inline int64_t maxCallGas(int64_t gas) {
 
       call_message.gas = gas;
 
+       ensureSenderBalance(call_message.value);
```
axic: I think we need to wrap this into `if (isZeroUint256(call_message.value)` or `if (import->base == Name("call") || import->base == Name("callCode"))` for explicitness.

PR #257, opened 2018-06-25
Fix storeMemory logic for extCodeCopy

lrettig:
Use offset 0 when reading from newly-created buffer. Otherwise we get out-of-bounds errors when we shouldn't.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -389,12 +389,12 @@ inline int64_t maxCallGas(int64_t gas) {
       takeInterfaceGas(GasSchedule::extcode + GasSchedule::copy * ((uint64_t(length) + 31) / 32));
 
       evmc_address address = loadUint160(addressOffset);
-      // FIXME: optimise this so not vector needs to be created
+      // FIXME: optimise this so no vector needs to be created
       vector<uint8_t> codeBuffer(length);
       size_t numCopied = context->fn_table->copy_code(context, &address, codeOffset, codeBuffer.data(), codeBuffer.size());
       fill_n(&codeBuffer[numCopied], length - numCopied, 0);
 
-      storeMemory(codeBuffer, codeOffset, resultOffset, length);
+      storeMemory(codeBuffer, 0, resultOffset, length);
 
       return Literal();
     }


src/eei.cpp
```diff
@@ -389,12 +389,12 @@ inline int64_t maxCallGas(int64_t gas) {
...
 
       evmc_address address = loadUint160(addressOffset);
-      // FIXME: optimise this so not vector needs to be created
+      // FIXME: optimise this so no vector needs to be created
       vector<uint8_t> codeBuffer(length);
       size_t numCopied = context->fn_table->copy_code(context, &address, codeOffset, codeBuffer.data(), codeBuffer.size());
       fill_n(&codeBuffer[numCopied], length - numCopied, 0);
 
-      storeMemory(codeBuffer, codeOffset, resultOffset, length);
+      storeMemory(codeBuffer, 0, resultOffset, length);
```
axic: This was missed in #172.

Is there a test case for this? Apparently not, otherwise current CI would fail.

lrettig: I'm writing one in https://github.com/ewasm/tests/pull/52, that's how I caught this. There was previously no test for `returnDataCopy`.

axic: But this is only used for `extCodeCopy` if I'm not mistaken? :)

lrettig:
I think that's correct

Sent via Superhuman ( https://sprh.mn/?vip=lane@ethereum.org )

src/eei.cpp
axic: To clarify: do we have a test case now to verify the code is correct after the change?

axic: I'm happy to merge this, but would prefer to have a test case for it.

lrettig: Correct that this is only for `extCodeCopy`. The test is in https://github.com/ewasm/tests/pull/52 which is blocking on https://github.com/ewasm/tests/pull/52#issuecomment-400667415 -- could use your feedback on that.

There's an [existing test](https://github.com/ewasm/tests/blob/wasm-tests/src/GeneralStateTestsFiller/stEWASMTests/extCodeCopyFiller.yml) for `extCodeCopy` but I think it didn't catch this because it uses an offset of zero. This [new (unmerged) test](https://github.com/lrettig/tests/blob/b75dd324d71db9dc0a57c422fd04292760287aad/src/GeneralStateTestsFiller/stEWASMTests/extCodeCopyOOBFiller.yml) caught it.

axic: > This new (unmerged) test caught it.

Can we merge this?

axic:
`codeOffset` is applied at `copy_code`, no need to do it again in `storeMemory`.

PR #266, opened 2018-07-24
 CMake: Add option to build with WAVM

chfast:
This adds -DHERA_WAVM option what for now will build WAVM and link hera with it. Because Hera does not uses WAVM it will not do anything useful, but we can check if the build works.

Also the `HERA_WAVM=1` macro is predefined for build with WAVM.

This depends on LLVM 6.0 so you have to install libllvm-6.0-dev. See http://apt.llvm.org.

The WAVM produces a number of libraries. In this PR only the libruntime is used, probably more is going to be needed when we start using WAVM.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -21,3 +21,8 @@ if(NOT WIN32)
     set_target_properties(hera PROPERTIES LINK_FLAGS "-Wl,-undefined,error")
   endif()
 endif()
+
+if(HERA_WAVM)
+    target_compile_definitions(hera PRIVATE HERA_WAVM=1)
+    target_link_libraries(hera PRIVATE wavm::wavm)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,6 +21,11 @@ project(hera)
 
 include(ProjectBinaryen)
 
+option(HERA_WAVM "Build with WAVM" OFF)
+if (HERA_WAVM)
+    include(ProjectWAVM)
+endif()
+
 add_subdirectory(evmc)
 add_subdirectory(evm2wasm)
 add_subdirectory(src)
--- a/README.md
+++ b/README.md
@@ -11,6 +11,13 @@ Currently it uses [Binaryen](https://github.com/webassembly/binaryen)'s interpre
 - `-DHERA_DEBUGGING=ON` will turn on debugging features and messages
 - `-DBUILD_SHARED_LIBS=ON` is a standard CMake option to build libraries as shared. This will build Hera shared library that can be then dynamically loaded by EVMC compatible Clients (e.g. `eth` from [cpp-ethereum]).
 
+### WAVM support
+
+*Unfinished support, work in progress.*
+
+- `-DHERA_WAVM=ON` will request the compilation of wavm support
+- `-DLLVM_DIR=...` one will need to specify the path to LLVM's CMake file. In most installations this has to be within the `lib/cmake/llvm` directory, such as `/usr/local/Cellar/llvm/6.0.1/lib/cmake/llvm` on Homebrew.
+
 ## Runtime options
 
 These are to be used via EVM-C `set_option`:
--- a/circle.yml
+++ b/circle.yml
@@ -201,9 +201,9 @@ jobs:
       - CC:  clang
       - GENERATOR: Ninja
       - BUILD_PARALLEL_JOBS: 4
-      - CMAKE_OPTIONS: -DBUILD_SHARED_LIBS=ON -DHERA_DEBUGGING=OFF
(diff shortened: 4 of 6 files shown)


chfast:
I think I should test the build with WAVM on any CI servers.

jakelang:
I think there is an existing option in the abstraction layer called `-DWAVM_SUPPORTED`. I will adjust to use this one instead.

jakelang:
@chfast the `linux-clang-shared-release` seems to be failing because WAVM depends on LLVM6.

src/CMakeLists.txt
```diff
@@ -22,3 +22,8 @@ if(NOT WIN32)
     set_target_properties(hera PROPERTIES LINK_FLAGS "-Wl,-undefined,error")
   endif()
 endif()
+
+if(HERA_WAVM)
+    target_compile_definitions(hera PRIVATE HERA_WAVM=1)
+    target_link_libraries(hera PRIVATE wavm::runtime)
```
jakelang: In addition to the Runtime lib (and the dependencies pulled in implicitly) we also need the WASM library to convert the bytecode into WAVM's internal IR.

chfast:
Ready to merge?

jakelang:
If the WAVM include root is added to Hera's include search list then yeah, I think it is ready.

axic:
Added a sample code to pull in all the required bits for loading & running (in https://github.com/ewasm/hera/pull/266/commits/dc507605c232101e09ec2069e1082bca79b53658) and it seems to compile.

axic:
My bad, used the wrong ifdef, however now it does complain:
```
In file included from /home/builder/project/src/hera.cpp:41:
In file included from deps/src/wavm/Include/IR/Module.h:3:
In file included from deps/src/wavm/Include/Inline/Assert.h:3:
deps/src/wavm/Include/Platform/Platform.h:89:2: error: unknown type name 'DLL_IMPORT'
        PLATFORM_API Uptr getPageSizeLog2();
        ^
deps/src/wavm/Include/Platform/Platform.h:31:23: note: expanded from macro 'PLATFORM_API'
        #define PLATFORM_API DLL_IMPORT
                             ^
deps/src/wavm/Include/Platform/Platform.h:89:19: error: expected ';' after top level declarator
        PLATFORM_API Uptr getPageSizeLog2();
                         ^
```

axic:
@chfast So it seems the sample code compiles and links, but there is some kind of runtime linking error:

```
testeth 1.4.0.dev0
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  boost::dll::shared_library::load() failed (dlerror system message: /home/builder/build/src/libhera.so: undefined symbol: _ZN4llvm17JITSymbolResolver6anchorEv): Bad file descriptor
Received 'aborted' signal
```

which resolves to `llvm::JITSymbolResolver::anchor()`

axic:
Lastly it fails to build if not built as shared library.

axic:
This seems to be working on dynamic loading now (tests fail because the code will throw intentionally).

Static linking however doesn't work because the flag `-Werror=maybe-uninitialized` is passed down from aleth.

axic:
The new sed doesn't work on mac:
```
[ 10%] Performing patch step for 'wavm'
sed: 1: "CMakeLists.txt": invalid command code C
```

axic:
It still fails on mac:
```
[ 10%] Performing patch step for 'wavm'
sed: 1: "CMakeLists.txt": invalid command code C
sed: 1: "CMakeLists.txt": invalid command code C
```

axic:
Moving them back to `-iE` fixes it.

chfast:
sed is still failing now?

axic:
Yes, changing your last commit to `sed -iE` makes everything to work.

axic:
Well, on this latest version I got a new error:
```
/Users/alex/Projects/hera/src/hera.h:27:58: error: expected function body after function declarator
EVMC_EXPORT struct evmc_instance* evmc_create_hera(void) EVMC_NOEXCEPT;
```

chfast:
But I removed regex from sed.

axic:
`sed` on Mac is kind of weird

axic:
On Mac, though `-iE` workarounds that by somehow making it to ignore the `extension` parameter:
```
     -i extension
             Edit files in-place, saving backups with the specified extension.  If a zero-length extension is given, no backup will be saved.  It is not recommended to give a zero-
             length extension when in-place editing files, as you risk corruption or partial content in situations where disk space is exhausted, etc.
```

chfast:
I will replace it with CMake script not to have problems with it on Windows.

axic:
> I will replace it with CMake script not to have problems with it on Windows.

I think I will merge this broken `sed -iE` version so we can get going and you can fix it in a new PR.

PR #273, opened 2018-07-29
Check for out of bounds source on memory copy

lrettig:
loadMemory was only checking for integer overflow/negative length. Also
check that we don't attempt to copy from beyond the end of memory. Also
perform the same check before attempting to resize a vector larger than
memory.

Tests in https://github.com/ewasm/tests/pull/57

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -490,6 +490,7 @@ inline int64_t maxCallGas(int64_t gas) {
         topics[i] = loadUint256(topicOffset);
       }
 
+      ensureSourceMemoryBounds(dataOffset, length);
       vector<uint8_t> data(length);
       loadMemory(dataOffset, data, length);
 
@@ -606,6 +607,7 @@ inline int64_t maxCallGas(int64_t gas) {
 
       HERA_DEBUG << (import->base == Name("revert") ? "revert " : "return ") << hex << offset << " " << size << dec << "\n";
 
+      ensureSourceMemoryBounds(offset, size);
       result.returnValue = vector<uint8_t>(size);
       loadMemory(offset, result.returnValue, size);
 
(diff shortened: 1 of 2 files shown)


lrettig:
@axic do you think `externalCodeCopy` is okay? [This line](https://github.com/lrettig/hera/blob/f1a750ee222dd9765a51b3d808e482957bf05408/src/eei.cpp#L396) in particular. No explicit check is made on `length` but gas is charged before the buffer is allocated.

src/eei.cpp
```diff
@@ -794,6 +803,9 @@ inline int64_t maxCallGas(int64_t gas) {
       // NOTE: this must be declared outside the condition to ensure the memory doesn't go out of scope
       vector<uint8_t> contract_code;
       if (length) {
+        // Perform this check here, rather than inside loadMemory, before we attempt to resize the vector
+        ensureCondition((dataOffset + length) >= dataOffset, InvalidMemoryAccess, "Out of bounds (source) memory copy.");
+        ensureCondition(memory.size() >= (dataOffset + length), InvalidMemoryAccess, "Out of bounds (source) memory copy.");
```
axic: Any way we can pull these into helpers? i'm not a fan of code duplication. (We let it pass with the gas calculation with the note that it will be refactored, but it wasn't.)

I suggest a helper like
```
ensureSourceBuffer(uint32_t offset, uint32_t length) {
    ensureCondition((offset + length) >= offset, InvalidMemoryAccess, "Out of bounds (source) memory copy.");
    ensureCondition(memory.size() >= (offset + length), InvalidMemoryAccess, "Out of bounds (source) memory copy.");
}
```

lrettig:
@axic back to you.

PR #281, opened 2018-07-31
Refactor ensureSenderBalance into enoughSenderBalanceFor

axic:
This simplifies the code significantly.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -742,17 +742,8 @@ inline int64_t maxCallGas(int64_t gas) {
 
       call_message.gas = gas;
 
-      // FIXME: consider refactoring this
       if (import->base == Name("call") || import->base == Name("callCode")) {
-        try {
-          ensureCondition(msg.depth < 1024, OutOfGas, "Call depth exceeded.");
-
-          ensureSenderBalance(call_message.value);
-        } catch (OutOfGas const&) {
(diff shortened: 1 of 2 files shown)


axic:
@chfast rebased, please review again.

src/eei.cpp
```diff
@@ -742,17 +742,8 @@ inline int64_t maxCallGas(int64_t gas) {
...
-        try {
-          ensureCondition(msg.depth < 1024, OutOfGas, "Call depth exceeded.");
-
-          ensureSenderBalance(call_message.value);
-        } catch (OutOfGas const&) {
-          // An out of gas error here doesn't really mean we're out of gas, it just means
-          // we don't have sufficient gas for the call value. We should return
-          // an error rather than throwing.
-          //
+        if (!enoughSenderBalanceFor(call_message.value) || (msg.depth >= 1024)) {
```
chfast: You should check the depth first, because it's quicker check.

axic: Right. I think actually we need to be more specific consensus wise and define which one is first.

chfast: The order does not affect EVM consensus because in any case the same failure is returned.

axic: In current EVM it is not an issue, but I think it is bad design not being specific of order.

Issue #157, opened 2018-03-14
Support wabt as a VM


axic:
Depends on #156.

axic:
@jakelang do you have a sample code to run instantiation with `wabt`?

@chfast can we have a `ProjectWABT`?

PR #283, opened 2018-07-31
Add wabt CMake file

axic:
Needs https://github.com/axic/wabt/tree/cmake-libs

Part of #157.


axic:
@chfast this isn't very useful but this is where I left if you want to pick it up

src/hera.cpp
```diff
@@ -29,9 +29,22 @@
...
+#include <src/option-parser.h>
+#include <src/resolve-names.h>
+#include <src/stream.h>
+#include <src/validator.h>
+#include <src/wast-parser.h>
+
 #include <evmc/evmc.h>
 
 #include <evm2wasm.h>
+#include <build/wabt/deps/src/wabt/src/string-view.h>
```
chfast: Remove it, my IDE added it automatically.

cmake/ProjectWabt.cmake
```diff
@@ -0,0 +1,43 @@
...
+    DOWNLOAD_DIR ${prefix}/downloads
+    SOURCE_DIR ${source_dir}
+    BINARY_DIR ${binary_dir}
+    URL https://github.com/WebAssembly/wabt/archive/1.0.5.tar.gz
+    URL_HASH SHA256=285700512a6af1524c16422d61ae4959d4b387f2a82698198eb524b514825a8a
+    CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
+    -DCMAKE_BUILD_TYPE=Release
+    -DBUILD_TESTS=OFF
+    -DCMAKE_INSTALL_LIBDIR=lib
```
axic: You can add `-DBUILD_TOOLS=OFF`, that change was merged upstream.

src/hera.cpp
```diff
@@ -29,6 +29,18 @@
...
+#include <src/common.h>
+#include <src/error-handler.h>
+#include <src/feature.h>
+#include <src/filenames.h>
+#include <src/ir.h>
+#include <src/option-parser.h>
+#include <src/resolve-names.h>
+#include <src/stream.h>
+#include <src/validator.h>
+#include <src/wast-parser.h>
```
axic: Any way to use a better path like `wabt/binary-writer.h` ?

chfast: Not without moving the files manually.

axic: And with patching upstream cmake?

PR #284, opened 2018-07-31
Introduce EEICallKind within the call* handler to simplify code

axic:
It also prepares it for VM abstraction.

Part of #156.


src/eei.h
```diff
@@ -62,6 +62,13 @@ struct EthereumInterface : ShellExternalInterface {
   }
 
 private:
+  enum EEICallKind {
```
chfast: Probably should be an `enum class`.

PR #285, opened 2018-07-31
Move setting ExecutionResult.gasLeft with startgas to its proper place

axic:


axic:
@chfast can you check this and raise a comment if something wasn't ok?

PR #286, opened 2018-07-31
Introduce hera_wasm_engine enum

axic:
Original code by @jakelang

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -41,6 +41,12 @@ using namespace std;
 using namespace wasm;
 using namespace HeraVM;
 
+enum hera_wasm_engine {
+  ENGINE_BINARYEN,
+  ENGINE_WAVM,
+  ENGINE_WABT
+};
+
 enum hera_evm_mode {
   EVM_REJECT,
   EVM_FALLBACK,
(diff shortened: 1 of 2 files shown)


src/hera.cpp
```diff
@@ -41,6 +41,12 @@ using namespace std;
 using namespace wasm;
 using namespace HeraVM;
 
+enum hera_wasm_engine {
```
chfast: 1. Why prefix `hera_` instead of `hera` namespace?
2. Prefer `enum class`.

axic: 1. I guess we can do a cleanup where we move it from `hera_` prefix to `hera` namespace (this applies to already existing code in there so I'd do it after this).

2. What is the difference/benefit of `enum class`?

chfast: We also use lower case names, leaving upper case only for macros. EVMC is an exception though.

jakelang:
Updated to use `class enum`

src/hera.cpp
```diff
@@ -496,6 +505,21 @@ int hera_set_option(
     hera->metering = strcmp(value, "true") == 0;
     return 1;
   }
+
+  if (strcmp(name, "engine") == 0) {
```
chfast: I prefer using `std::string` for comparisons, but I'm ok with leaving this as is because of the previous comparisons using `stdcmp()`.

jakelang: @chfast I plan to refactor this in a separate PR.

PR #287, opened 2018-07-31
getBlockhash() to signal success or failure

axic:
Implements https://github.com/ewasm/design/pull/76. Closes #111.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -274,9 +274,13 @@ inline int64_t maxCallGas(int64_t gas) {
 
       takeInterfaceGas(GasSchedule::blockhash);
       context->fn_table->get_block_hash(&blockhash, context, number);
+
+      if (isZeroUint256(blockhash))
+        return Literal(uint32_t(1));
+
       storeUint256(blockhash, resultOffset);
 
-      return Literal();
+      return Literal(uint32_t(0));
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -274,9 +274,13 @@ inline int64_t maxCallGas(int64_t gas) {
 
       takeInterfaceGas(GasSchedule::blockhash);
       context->fn_table->get_block_hash(&blockhash, context, number);
+
+      if (isZeroUint256(blockhash))
+        return Literal(static_cast<uint32_t>(1));
```
chfast: If you want a precise type for a literal, you can write `uint32_t(1)`.
I think it would be better to use signed type for error codes.

axic: > I think it would be better to use signed type for error codes.

The underlying binaryen type is `uint32_t`. In wasm the stack item's interpretation depend on the opcode (just like in evm).

chfast: That's not what I meant. You don't need to use `static_cast` for literals, because in some sense it's compile-time conversion. Because casts might be a cause of some problems, it's better to avoid the "unsafe" form here by writing equivalent `uint32_t(1)`. This is also equivalent of `1u`.

PR #289, opened 2018-08-03
Add option to preload system contracts from file before execution

### LINKED ISSUE #288: Add commandline option to load bytecode for system contracts
Proposed syntax:
- `sys:<address>=<path>`
- `sys:sentinel=<path>` (alias to 0x00...00e?)
- `sys:evm2wasm=<path>` (likewise)

For example, one would run `testeth -t GeneralStateTests/stExample -- --testpath ~/tests --singlenet "Byzantium" --singletest "add11" --vm ~/libhera.so --evmc sys:evm2wasm=my_cool_version_of_evm2wasm.wasm`

`sys` stands for system contract, while the `address` can be any ethereum address (hex encoded, all nibbles set, with the `0x` prefix) and `path` is a relative filesystem path to load the binary from.

Additionally `sentinel` and `evm2wasm` are an aliases for their respective addresses.

The main benefit here is that we can load system contracts without adjusting the genesis and as a result makes testing way easier.

In the future we could think about migrating all these settings over to a configuration file.

jakelang:
Closes #288 
Syntax: `--evmc sys:<address or alias>=<filepath>`
Hex address must be formatted with a leading `0x` and must be 40 characters long (all nibbles must be explicitly specified, including leading zeroes).

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -83,6 +83,7 @@ struct hera_instance : evmc_instance {
   std::unique_ptr<WasmEngine> engine{new BinaryenEngine};
   hera_evm_mode evm_mode = hera_evm_mode::reject;
   bool metering = false;
+  vector<pair<evmc_address, string>> contract_preload_list;
 
   hera_instance() noexcept : evmc_instance({EVMC_ABI_VERSION, "hera", hera_get_buildinfo()->project_version, nullptr, nullptr, nullptr, nullptr}) {}
 };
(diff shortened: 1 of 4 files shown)


src/hera.cpp
```diff
@@ -73,6 +74,40 @@ bool hasWasmPreamble(vector<uint8_t> const& _input) {
     _input[7] == 0;
 }
 
+bool deploySystemContract(string & code, evmc_address addr) {
+  evmc_message create_message;
+  create_message.input_data = reinterpret_cast<const uint8_t *>(code.c_str());
+  create_message.destination = addr;
+
+  //Prepare message call to deploy system contract
```
jakelang: actually it might make sense to just create a system contract map (addresses to function pointers)

jakelang:
Rebased.

jakelang:
@axic updated this to override the code at runtime, without CREATEing system contracts

src/hera.cpp
```diff
@@ -488,6 +588,109 @@ evmc_result hera_execute(
...
+
+  pair<evmc_address, bool> ret = { {}, false };
+  string evmc_option_raw = string(name);
+
+#if HERA_DEBUGGING
+  cerr << "Trying to parse EVMC option as preload flag: " << evmc_option_raw << endl;
+#endif 
+
+  //Check the "sys:" syntax by comparing substring
+  if (evmc_option_raw.length() < 4 || evmc_option_raw.substr(0, 4).compare("sys:") != 0) {
```
axic: Canno you just use `evmc_option_raw.find("sys:") == 0` here?

src/hera.cpp
```diff
@@ -61,6 +61,7 @@ struct hera_instance : evmc_instance {
   hera_wasm_engine wasm_engine = hera_wasm_engine::binaryen;
   hera_evm_mode evm_mode = hera_evm_mode::reject;
   bool metering = false;
+  vector<pair<evmc_address, string>> contract_preload_list;
```
axic: Why not use a `map` here?

src/hera.cpp
```diff
@@ -488,6 +588,109 @@ evmc_result hera_execute(
...
+
+  pair<evmc_address, bool> ret = { {}, false };
+  string evmc_option_raw = string(name);
+
+#if HERA_DEBUGGING
+  cerr << "Trying to parse EVMC option as preload flag: " << evmc_option_raw << endl;
+#endif 
+
+  //Check the "sys:" syntax by comparing substring
+  if (evmc_option_raw.length() < 4 || evmc_option_raw.find("sys:") != 0) {
```
axic: I think there isn't a need to check the length if using `find`.

src/hera.cpp
```diff
@@ -488,6 +588,109 @@ evmc_result hera_execute(
...
+  }
+
+#if HERA_DEBUGGING
+  cerr << "Successfully unmarshalled hex string into address struct" << endl;
+#endif
+
+  return pair<evmc_address, bool>(ret, true);
+}
+
+pair<evmc_address, bool> parse_preload_addr(const char *name)
```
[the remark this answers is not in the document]
axic: `address` is a string and can contain an alias or a hex string.

src/hera.cpp
```diff
@@ -64,6 +64,7 @@ struct hera_instance : evmc_instance {
   hera_wasm_engine wasm_engine = hera_wasm_engine::binaryen;
   hera_evm_mode evm_mode = hera_evm_mode::reject;
   bool metering = false;
+  vector<pair<evmc_address, string>> contract_preload_list;
```
axic: I think this should be `vector<pair<evmc_address, vector<uint8_t>>>` so that the at runtime it doesn't need to be loaded every time.

PR #292, opened 2018-08-09
Refactor call* handling for readability

axic:
Part of #156.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -675,6 +675,16 @@ inline int64_t maxCallGas(int64_t gas) {
       uint32_t dataOffset;
       uint32_t dataLength;
 
+      if (kind == EEICallKind::Call || kind == EEICallKind::CallCode) {
+        valueOffset = arguments[2].geti32();
+        dataOffset = arguments[3].geti32();
+        dataLength = arguments[4].geti32();
+      } else {
+        valueOffset = 0;
+        dataOffset = arguments[2].geti32();
+        dataLength = arguments[3].geti32();
+      }
+
       heraAssert((msg.flags & ~EVMC_STATIC) == 0, "Unknown flags not supported.");
 
       evmc_message call_message;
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -671,6 +671,16 @@ inline int64_t maxCallGas(int64_t gas) {
       uint32_t dataOffset;
       uint32_t dataLength;
 
+      if (kind == EEICallKind::Call || kind == EEICallKind::CallCode) {
+        valueOffset = arguments[2].geti32();
+        dataOffset = arguments[3].geti32();
+        dataLength = arguments[4].geti32();
+      } else {
+        valueOffset = 0;
```
chfast: It's better to always init variables. Then this is not needed.

src/eei.cpp
```diff
@@ -679,33 +689,28 @@ inline int64_t maxCallGas(int64_t gas) {
...
-        }
+        break;
+      case EEICallKind::CallDelegate:
+        call_message.sender = msg.sender;
+        call_message.value = msg.value;
+        call_message.kind = EVMC_DELEGATECALL;
+        break;
+      case EEICallKind::CallStatic:
+        call_message.sender = msg.destination;
+        call_message.value = {};
```
chfast: This is not needed. The `.value` is already zerod.

src/eei.cpp
```diff
@@ -679,33 +689,28 @@ inline int64_t maxCallGas(int64_t gas) {
...
+        call_message.kind = EVMC_CALL;
```
chfast: Just for clarity, I'd do the `.kind` assignment as the first one.

src/eei.cpp
axic: True, although they are implicitly initialised to 0 anyway. Here they are grouped, and I think it makes it clear that in this case value is ignored as option.

src/eei.cpp
axic: I think these are done allover the code because at some point you wanted them to be explicitly initialised?

Happy to do a cleanup, but would do it in a separate PR, since there are about 20 lines affected.

src/eei.cpp
axic: Will do it.

src/eei.cpp
chfast: The `valueOffset` is not initialized. I'm ok with leaving this assignment here.

PR #294, opened 2018-08-09
Extract EEI methods from within callImport (call, create, selfDestruct, returnDataCopy)

axic:
Part of #156.

Inspired by #148.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -596,11 +596,7 @@ namespace hera {
     if (import->base == Name("getReturnDataSize")) {
       heraAssert(arguments.size() == 0, string("Argument count mismatch in: ") + import->base.str);
 
-      HERA_DEBUG << "getReturnDataSize\n";
-
-      takeInterfaceGas(GasSchedule::base);
-
-      return Literal(static_cast<uint32_t>(m_lastReturnData.size()));
+      return Literal(eeiGetReturnDataSize());
     }
 
     if (import->base == Name("returnDataCopy")) {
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -722,8 +755,18 @@ inline int64_t maxCallGas(int64_t gas) {
...
+      case EEICallKind::Call: methodName = "call"; break;
+      case EEICallKind::CallCode: methodName = "callCode"; break;
+      case EEICallKind::CallDelegate: methodName = "callDelegate"; break;
+      case EEICallKind::CallStatic: methodName = "callStatic"; break;
+      }
+#endif
+
       HERA_DEBUG <<
-        import->base << " " << hex <<
+        methodName << " " << hex <<
```
jakelang: this needs to be conditionally compiled with HERA_DEBUG for release builds to work correctly.

axic: That's fine, the main question in this PR is the naming and general idea.

src/eei.cpp
```diff
@@ -659,6 +657,40 @@ namespace hera {
...
+    heraAssert(false, string("Unsupported import called: ") + import->module.str + "::" + import->base.str + " (" + to_string(arguments.size()) + "arguments)");
+  }
+
+  void EthereumInterface::eeiReturnDataCopy(uint32_t dataOffset, uint32_t offset, uint32_t size)
+  {
+      safeChargeDataCopy(size, GasSchedule::verylow);
+      storeMemory(m_lastReturnData, offset, dataOffset, size);
+  }
+
+  uint32_t EthereumInterface::eeiCall(EEICallKind kind, int64_t gas, uint32_t addressOffset, uint32_t valueOffset, uint32_t dataOffset, uint32_t dataLength)
```
jakelang: I think it may make sense to trade some code repetition for more clarity here, as in #148 where they are split.

axic: That was the main point of enum so that it can be shared 😉

src/eei.h
```diff
@@ -76,6 +76,8 @@ struct EthereumInterface : ShellExternalInterface {
 
   // EEI methods
 
+  uint32_t eeiGetReturnDataSize();
```
jakelang: Refer to #148 for the ordering of these function declarations. I think from a maintainability perspective it makes sense to have a more logical ordering of the methods (the design doc needs to also be re-ordered).

axic: Yeah, would need to apply that to both design, hera, ewasm-rust-api, as-ewasm-api.

But perhaps not on this PR.

PR #298, opened 2018-08-10
Upgrade EVMC to v5.0.0

chfast:


axic:
Need to update circleci with an evmc5 aleth.

axic:
Need to update the static build too. Please use the very same hash as the prebuilt binary is built from.

PR #299, opened 2018-08-10
Refactor topic handling in log()

axic:
Part of #156.

Inspired by #148.

Helps with #294.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -481,6 +481,10 @@ inline int64_t maxCallGas(int64_t gas) {
       uint32_t dataOffset = arguments[0].geti32();
       uint32_t length = arguments[1].geti32();
       uint32_t numberOfTopics = arguments[2].geti32();
+      uint32_t topic1 = arguments[3].geti32();
+      uint32_t topic2 = arguments[4].geti32();
+      uint32_t topic3 = arguments[5].geti32();
+      uint32_t topic4 = arguments[6].geti32();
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -481,18 +481,23 @@ inline int64_t maxCallGas(int64_t gas) {
...
+      uint32_t topic3 = arguments[5].geti32();
+      uint32_t topic4 = arguments[6].geti32();
 
       HERA_DEBUG << "log " << hex << dataOffset << " " << length << " " << numberOfTopics << dec << "\n";
 
       ensureCondition(!(msg.flags & EVMC_STATIC), StaticModeViolation, "log");
 
       ensureCondition(numberOfTopics <= 4, ContractValidationFailure, "Too many topics specified");
 
+      // FIXME: should this assert that unused topic offsets must be 0?
```
axic: Created an issue: https://github.com/ewasm/design/issues/129

chfast:
I'm not sure where it is going.

axic:
> I'm not sure where it is going.

See #294.

These will be moved into their individual methods. Then `EthereumInterface` will be a class of its own without any binaryen influence and all the VMs will have their own subclass inheriting from it, implementing the `memory*` helpers (will be virtual then) and mapping the `eii*` functions in their respective host function wrappers.

chfast:
Ok, so `topicX` are going to be passed to a function, not abstracting only memory access.

axic:
Correct, the host function wrappers need to acquire the arguments as i32/i64 pass down to the actual implementation and return i32/i6.

PR #301, opened 2018-08-10
Add specific exception for wasm VM traps

axic:


src/exceptions.h
```diff
@@ -29,6 +29,9 @@ class HeraException : public std::exception {
 class InternalErrorException : public HeraException {
   using HeraException::HeraException;
 };
+class VMTrap : public HeraException {
```
axic: Any better name suggestion here?

PR #302, opened 2018-08-10
Add specific exception for out of range parameters

axic:


axic:
As a second look, I think these should be internal errors. These are used to read value from EVMC and ensure they are 128 bit wide.

What is the best way to handle this?

axic:
So basically these two cases can happen if the client is not "ewasm compatible", that means an account has a balance larger than 128 bits.

@chfast @jakelang any idea how to name it?

chfast:
From C++:
- https://en.cppreference.com/w/cpp/error/out_of_range
- https://en.cppreference.com/w/cpp/error/invalid_argument
From Java:
- https://docs.oracle.com/javase/7/docs/api/java/lang/IllegalArgumentException.html

PR #304, opened 2018-08-10
Circle CI: Use Aleth 1.4.0rc1 for testing

chfast:
Why is the debug version of shared library failing?


chfast:
> I guess the d suffix is not handled by evmc-loader ?

I forgot it guesses the function name by the library name, so it would guess "herad".

axic:
Please update the hash key:
>        key: &eth-cache-key cpp-prebuilt-cache-{{arch}}-{{checksum "compiler.version"}}-2ddd9ffb1d91b138b81716e343affaeae74dc645

axic:
There's some build issue.

PR #310, opened 2018-08-14
Code Coverage

### LINKED ISSUE #309: Set up codecov for Hera

chfast:
Closes #309

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,13 +1,13 @@
 cmake_minimum_required(VERSION 3.5)
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+include(cmake/cable/bootstrap.cmake)
+include(CableToolchains)
+include(CableCompilerSettings)
 
-set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/cmake/toolchains/cxx11-pic.cmake CACHE FILEPATH "CMake toolchain file")
+cable_configure_toolchain(DEFAULT cxx11-pic)
 
 set(CMAKE_DEBUG_POSTFIX "")
 set(HUNTER_CONFIGURATION_TYPES Release CACHE STRING "Build type of the 3rd party packages")
-
-add_compile_options(-Wall -Werror -Wextra -Wconversion -Wno-sign-conversion -Wno-unknown-pragmas)
 
 if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/evm2wasm/cmake/HunterGate.cmake)
     include(evm2wasm/cmake/HunterGate.cmake)
@@ -18,6 +18,10 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/evm2wasm/cmake/HunterGate.cmake)
 endif()
 
 project(hera)
+
+cable_configure_compiler()
+# TODO: fix the warnings instead
+add_compile_options(-Wno-pedantic -Wno-sign-conversion)
 
 include(ProjectBinaryen)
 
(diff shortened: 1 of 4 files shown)


chfast:
Looks like the evm2wasm build get messed up when building inside Aleth. I'm not sure why. Probably the easiest approach would be to enabled `-Wsign-conversion` in evm2wasm and fix all warnings.

CMakeLists.txt
```diff
@@ -1,14 +1,14 @@
 cmake_minimum_required(VERSION 3.5)
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+include(cmake/cable/bootstrap.cmake)
+include(CableToolchains)
+include(CableCompilerSettings)
 
-set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/cmake/toolchains/cxx11-pic.cmake CACHE FILEPATH "CMake toolchain file")
```
axic: Ned to remove the extra toolchain file then.

CMakeLists.txt
```diff
@@ -19,6 +19,9 @@ endif()
 
 project(hera)
 
+cable_configure_compiler()
+add_compile_options(-Wno-pedantic -Wno-sign-conversion)
```
axic: Does cable adds `-Wpedantic` so that we need to negative it here?

Any point keeping `-Wno-unknown-pragmas`?

axic: Can you remove `-Wno-sign-conversion` and update `evm2wasm`?

chfast: The `-Wno-sign-conversion` is also needed in Hera.

chfast:
This is ready.

CMakeLists.txt
axic: Tracked in #311.

CMakeLists.txt
```diff
@@ -1,14 +1,14 @@
 cmake_minimum_required(VERSION 3.5)
 
-list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
+include(cmake/cable/bootstrap.cmake)
+include(CableToolchains)
+include(CableCompilerSettings)
 
-set(CMAKE_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/cmake/toolchains/cxx11-pic.cmake CACHE FILEPATH "CMake toolchain file")
+cable_configure_toolchain(DEFAULT cxx11-pic)
```
axic: Isn't this getting the toolchain from cable instead? Seems not since I've removed it from the `cmake/toolchains` directory and it is failing now.

circle.yml
```diff
@@ -14,7 +14,8 @@ defaults:
         echo Generator: $GENERATOR
         echo CMake options: $CMAKE_OPTIONS
         $CXX --version
-        $CXX --version > compiler.version
+        $CXX --version >> toolchain
```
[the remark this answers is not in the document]
chfast: It could be, but this a CI and we start from clean environment.

CMakeLists.txt
chfast: I  taught Cable to also check `${PROJECT_SOURCE_DIR}/cmake/toolchains` dir.

circle.yml
```diff
@@ -224,14 +230,14 @@ jobs:
       - *fetch-tests
       - *test
 
-  linux-clang-shared-debug:
+  linux-gcc-shared-coverage:
```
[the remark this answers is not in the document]
chfast: 1. coverage implies build with debugging symbols.
2. Checking coverage in release builds is not doable. Let's have a base first, then we can start tweaking the config. There are other ways of dealing with such issues like blacklisting some files.
3. Also, I don't think we have a capacity to run debug and coverage builds at the moment.

chfast:
@axic, what have you done that it is failing now?

PR #312, opened 2018-08-15
Fix warnings for signedness

### LINKED ISSUE #311: Fix warnings
Enable `-Wsign-conversions` and `-padantic` and fix all the warnings that the compiler will raise.

axic:
Part of #311.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -62,7 +62,7 @@ namespace hera {
     if (import->base == Name("print32")) {
       heraAssert(arguments.size() == 1, string("Argument count mismatch in: ") + import->base.str);
 
-      uint32_t value = arguments[0].geti32();
+      uint32_t value = static_cast<uint32_t>(arguments[0].geti32());
 
       cerr << "DEBUG print32: " << value << " " << hex << "0x" << value << dec << endl;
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -370,20 +370,20 @@ inline int64_t maxCallGas(int64_t gas) {
 
       takeInterfaceGas(GasSchedule::base);
 
-      return Literal(static_cast<uint32_t>(m_code.size()));
+      return Literal(uint32_t(m_code.size()));
```
chfast: Do not hide a cast.

axic:
> Unsigned types should be avoided. They are only good for bitfields and bit manipulation.

I don't get it, why?

axic:
Still, why?

chfast:
I had to look it up. Most of the arguments are listed here (read all section): https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#arithmetic.

I should add that you can detect signed overflow with Undefined Behavior Sanitizer (with the newest version you can also detect unsigned overflows, but because they are valid in C++, you can get false positives). Together with fuzz testing this gives incredible results.

axic:
As discussed the specification isn't clear what the types of these are, but implicitly it feels they are defined as unsigned in the EEI. Added an issue to review that: https://github.com/ewasm/design/issues/132

Merging this change as it doesn't changes semantics.

Issue #318, opened 2018-08-16
Clarify endianness of memory access operations


jakelang:
It has become a bit confusing that the overloaded functions for memory access actually have different functionality, in that some automatically convert endianness and some do not.

axic:
So basically Hera considers both 128, 160 and 256bit values and numbers and does byteswapping on them.

However, only 128-bit should be a number given that refers to "Ether value", while 160-bit values represent addresses and 256-bit values represent storage keys, values and log data.

PR #319, opened 2018-08-16
Add 'Reverse' suffix to loadMemory/storeMemory with byte pointer

axic:
Part of #318.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -920,7 +920,7 @@ inline int64_t maxCallGas(int64_t gas) {
     ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Out of bounds (source) memory copy.");
   }
 
-  void EthereumInterface::loadMemory(uint32_t srcOffset, uint8_t *dst, size_t length)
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -920,7 +920,7 @@ inline int64_t maxCallGas(int64_t gas) {
     ensureCondition(memorySize() >= (offset + length), InvalidMemoryAccess, "Out of bounds (source) memory copy.");
   }
 
-  void EthereumInterface::loadMemory(uint32_t srcOffset, uint8_t *dst, size_t length)
+  void EthereumInterface::loadMemoryReverse(uint32_t srcOffset, uint8_t *dst, size_t length)
```
jakelang: Perhaps it would make sense to explicitly annotate the function name, e.g. BE2LE
Or do the c++ thing and have an `IOManager` class wrapping around the ShellInterface with templated functions for performing the access operations with or without endianness swapping.

In general, the data flow in and out of EVMC and the WASM module could be better represented with another abstraction layer.

axic: Well this is the first step to make it apparent. Then I think we need to review it and ensure that byteswapping only happens on numeric types (practically only balance). Storage should not be swapped.

PR #320, opened 2018-08-16
Public header and CMake package

### LINKED ISSUE #300: Move public hera.h header into include/hera.h

chfast:
Resolves #300.


chfast:
If we want to proceed with this now, I would need to include a fix in Aleth 1.4.0rc2.

axic:
> If we want to proceed with this now, I would need to include a fix in Aleth 1.4.0rc2.

For what exactly?

axic:
Also can we merge the evm2wasm change without merging this here?

chfast:
>> If we want to proceed with this now, I would need to include a fix in Aleth 1.4.0rc2.
> 
> For what exactly?

Aleth does not build because we moved the hera header file.


> Also can we merge the evm2wasm change without merging this here?

Yes.

axic:
Please use the hash of master in evm2wasm (4e6a41c10c742694eeceb92cd0e73e4108f812d8).

CMakeLists.txt
```diff
@@ -18,6 +20,7 @@ if(EXISTS ${CMAKE_CURRENT_SOURCE_DIR}/evm2wasm/cmake/HunterGate.cmake)
 endif()
 
 project(hera)
+set(PROJECT_VERSION 0.1.0)
```
axic: I feel this should be done its own PR, adding proper build info. Is this line needed for this PR? If not, please remove. Otherwise please create a buildinfo PR first.

chfast: This must be here, because CMake needs the version specified for making the package. I just put any number there. Can be also `0.0.0`.

axic:
Updated the evm2wasm commit, but didn't change anything.

Need to be rebased.

chfast:
Do not rebase.

chfast:
This is ready. Can we merge it as is? Aleth now references one of the commits from this PR.

PR #321, opened 2018-08-16
Add safe helper for charging gas based on word count

jakelang:


src/eei.cpp
```diff
@@ -1056,4 +1044,11 @@ inline int64_t maxCallGas(int64_t gas) {
     }
     return true;
   }
+
+  void EthereumInterface::safeChargeDataCopy(uint32_t length, bool isExternalCode) {
+    static_assert(GasSchedule::copy <= 3, "Gas charge value has changed!");
+    takeGas((isExternalCode) ? GasSchedule::extcode : GasSchedule::verylow);
```
axic: I would just have a `baseCost` or `baseGas` field instead.

axic:
Also need to refactor the code in `log`.

jakelang:
@axic log wouldn't make sense to refactor because the helper would only be used once

axic:
#323 was meged, this needs to be rebased.

src/eei.cpp
```diff
@@ -302,13 +302,7 @@ inline int64_t maxCallGas(int64_t gas) {
...
       HERA_DEBUG << "callDataCopy " << hex << resultOffset << " " << dataOffset << " " << length << dec << "\n";
 
-      ensureCondition(ffs(GasSchedule::copy) + (ffsl(length) - 5) <= 64, OutOfGas, "Gas charge overflow");
-      ensureCondition(
-        numeric_limits<uint64_t>::max() - GasSchedule::verylow >= GasSchedule::copy * ((uint64_t(length) + 31) / 32),
-        OutOfGas,
-        "Gas charge overflow"
-      );
-      takeInterfaceGas(GasSchedule::verylow + GasSchedule::copy * ((uint64_t(length) + 31) / 32));
+      EthereumInterface::safeChargeDataCopy(length, GasSchedule::verylow);
```
axic: Don't need `EthereumInterface` here.

src/eei.h
```diff
@@ -107,6 +107,8 @@ struct EthereumInterface : ShellExternalInterface {
   /* Checks if 256 bit value is all zeroes */
   static bool isZeroUint256(evmc_uint256be const& value);
 
+  void safeChargeDataCopy(uint32_t length, uint32_t basecost);
```
axic: Please add a comment describing what it does.

src/eei.cpp
```diff
@@ -1056,4 +1044,12 @@ namespace hera {
     }
     return true;
   }
+
+  //Safely charges gas for callDataCopy, returnDataCopy, codeCopy, and externalCodeCopy
```
axic: I meant to put the short description into the header and it should only say something along the lines of "safely (handling overflows) charge variable gas cost based on word size"

src/eei.cpp
```diff
@@ -1056,4 +1044,12 @@ namespace hera {
...
+  void EthereumInterface::safeChargeDataCopy(uint32_t length, uint32_t basecost) {
```
axic: Please still change `basecost` to `baseCost`.

axic: Need to also add a `static_assert` for `baseCost`.

PR #322, opened 2018-08-16
Do not consider addresses as numbers (do not change byteorder)

axic:
Part of #318.

Part of https://github.com/ewasm/design/pull/105


src/eei.cpp
```diff
@@ -962,6 +976,19 @@ inline int64_t maxCallGas(int64_t gas) {
     }
   }
 
+  void EthereumInterface::storeMemory(const uint8_t *src, uint32_t dstOffset, uint32_t length)
```
axic: We should refactor this **after* this PR.

PR #323, opened 2018-08-16
Fix EVM compatibility in gas charges for ReturnDataCopy

jakelang:
https://ethereum.github.io/yellowpaper/paper.pdf - appendix H shows that the cost for returndatacopy has changed to be more in the style of *copy.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -644,7 +644,7 @@ inline int64_t maxCallGas(int64_t gas) {
 
       HERA_DEBUG << "returnDataCopy " << hex << dataOffset << " " << offset << " " << size << dec << "\n";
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -644,7 +643,7 @@ inline int64_t maxCallGas(int64_t gas) {
 
       HERA_DEBUG << "returnDataCopy " << hex << dataOffset << " " << offset << " " << size << dec << "\n";
 
-      takeInterfaceGas(GasSchedule::verylow);
+      takeInterfaceGas(GasSchedule::verylow + GasSchedule::copy * ((size + 31) / 32));
```
axic: Can you verify this in aleth?

axic: Checked, it is the case apparently.

PR #328, opened 2018-08-18
CircleCI: remove ewasm-tests because it is covered by the linux tests already

axic:


circle.yml
```diff
@@ -206,6 +168,7 @@ jobs:
       - *save-eth-cache
       - *test-evm2wasm
 
+  # to run this using local circleci tool, rename Ewasm-Tests to `build` then do `circleci build -c circle.yml`
```
chfast: Is this accurate when `ewasm-tests` is removed?

axic: I don't have the `circleci` commandline tool on the mac, not sure how it works and why cannot it point to a workflow task (instead of build)?

chfast: Remove the comment. Seems obsolete.

axic:
@chfast updated.

PR #329, opened 2018-08-18
CircleCI: run evm2wasm tests with shared Hera

### LINKED ISSUE #305: CI: Test evm2wasm.js with Aleth binaries
Convert the CircleCI job "evm2wasm-tests" to be like "linux-clang-shared-*" (or include the tests in this job), i.e. build hera to a shared library and use downloaded binaries of Aleth instead of building it from source.

axic:
Fixes #305.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -103,20 +103,6 @@ defaults:
         cd hera
         echo "hera branch: $(git rev-parse --abbrev-ref HEAD)  commit: $(git rev-parse HEAD)"
 
-  test-evm2wasm: &test-evm2wasm
-    run:
-      name: "Run evm2wasm state tests"
-      working_directory: ~/project
-      command: |
-        echo "print working directory:"
-        pwd
-        echo "list files in working directory:"
-        ls -al .
-        echo "check testeth command"
-        testeth --help
-        echo "run tests script."
-        ./scripts/evm2wasm-tests.sh
-
   install-cpp-ethereum: &install-cpp-ethereum
     run:
       name: "Install cpp-ethereum"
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -275,12 +216,35 @@ jobs:
       - *build
       - *save-deps-cache
 
+  linux-shared-evm2wasm:
```
chfast: Don't you want to merge this tests with `linux-*-shared-release` or `linux-*-shared-coverage`?

axic:
> Does this include tests with evm2wasm.js?

We do not use evm2wasm.js anymore. It is too slow.

circle.yml
axic: I don't want to include it in coverage.

chfast: So add it to  linux-clang-shared-release

axic:
@chfast rebased, can you take this over?

chfast:
Yes

circle.yml
```diff
@@ -137,6 +123,20 @@ defaults:
         testeth --version
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium
 
+  fetch-evm2wasm-tests: &fetch-evm2wasm-tests
+    run:
+      name: "Fetch evm2wasm tests"
+      command: |
+        git clone https://github.com/ethereum/tests
```
axic: This is wrong, should clone `ewasm/tests` with `ewasm-tests` branch.

chfast:
I will do it today.

chfast:
Included in `linux-clang-shared-release` but fails on some tests.

axic:
Can you just use the regular tests repo for evm2wasm? Don't want to spend any time on this.

chfast:
Sure.

chfast:
No clue why it does not work.

axic:
> No clue why it does not work.

I think it may be the entire problem we're having, that `aleth 1.4.x` is somehow broken.

The only difference in this PR is that the old version runs off master of aleth.

chfast:
Ok, it works now.

chfast:
Can we merge this?

PR #332, opened 2018-08-19
Clean up the handling of gas (use int64_t only)

axic:
Part of #311.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -200,9 +200,11 @@ namespace hera {
     if (import->base == Name("useGas")) {
       heraAssert(arguments.size() == 1, string("Argument count mismatch in: ") + import->base.str);
 
-      uint64_t gas = static_cast<uint64_t>(arguments[0].geti64());
+      int64_t gas = arguments[0].geti64();
 
       HERA_DEBUG << "useGas " << gas << "\n";
+
+      ensureCondition(gas >= 0, ArgumentOutOfRange, "Negative gas supplied.");
 
       takeGas(gas);
 
@@ -214,7 +216,7 @@ namespace hera {
 
       HERA_DEBUG << "getGasLeft\n";
 
-      static_assert(is_same<decltype(m_result.gasLeft), uint64_t>::value, "uint64_t type expected");
+      static_assert(is_same<decltype(m_result.gasLeft), int64_t>::value, "int64_t type expected");
 
       takeInterfaceGas(GasSchedule::base);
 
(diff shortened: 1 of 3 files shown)


axic:
@chfast I think this should be correct now (tests are failing though due to the edge cases I guess?). Can you have a look at the logic in this PR?

src/eei.cpp
```diff
@@ -485,7 +487,7 @@ namespace hera {
       static_assert(GasSchedule::logTopic <= 65536, "Gas cost of logTopic could lead to overflow");
       static_assert(GasSchedule::logData <= 65536, "Gas cost of logData could lead to overflow");
       // Using uint64_t to force a type issue if the underlying API changes.
-      takeInterfaceGas(GasSchedule::log + (GasSchedule::logTopic * numberOfTopics) + (GasSchedule::logData * static_cast<uint64_t>(length)));
+      takeInterfaceGas(GasSchedule::log + (GasSchedule::logTopic * numberOfTopics) + (GasSchedule::logData * static_cast<int64_t>(length)));
```
chfast: This is safe cast, so you can use `int64_t(length)`.

src/eei.cpp
```diff
@@ -1012,7 +1018,7 @@ namespace hera {
     // Allow 16 bits here.
     static_assert(GasSchedule::copy <= 65536, "Gas cost of copy could lead to overflow");
     // Using uint64_t to force a type issue if the underlying API changes.
-    takeInterfaceGas(GasSchedule::copy * ((static_cast<uint64_t>(length) + 31) / 32));
+    takeInterfaceGas(GasSchedule::copy * ((static_cast<int64_t>(length) + 31) / 32));
```
chfast: This cast is safe, could use `int64_t(length)`.

src/eei.cpp
```diff
@@ -1001,7 +1007,7 @@ namespace hera {
   /*
    * Utilities
    */
-  void EthereumInterface::safeChargeDataCopy(uint32_t length, uint32_t baseCost) {
+  void EthereumInterface::safeChargeDataCopy(uint32_t length, unsigned baseCost) {
```
chfast: Why have you changed this?
From my little experience I recommend using `int` or `int64_t` for cost constants as well.

axic: The constants are currently defined as `unsigned`, wanted to ensure it is the same here.

chfast: :+1:

PR #333, opened 2018-08-19
Add ArgumentOutOfRange error code

axic:


src/exceptions.h
```diff
@@ -32,6 +32,9 @@ class InternalErrorException : public HeraException {
 class VMTrap : public HeraException {
   using HeraException::HeraException;
 };
+class ArgumentOutOfRange : public HeraException {
```
jakelang: Is this intended to be used in OOB exceptions for the EEI?

axic: More like out of range for value types, such as `blockNumber` too high for `getBlockHash`, `gas` negative for `useGas`, etc.

src/hera.cpp
```diff
@@ -450,6 +450,12 @@ evmc_result hera_execute(
     ret.status_code = EVMC_FAILURE;
 #if HERA_DEBUGGING
     cerr << e.what() << endl;
+#endif
+  } catch (ArgumentOutOfRange const& e) {
+    // TODO: use specific error code? EVMC_ARGUMENT_OUT_OF_RANGE?
```
axic: Added a PR to EVMC: https://github.com/ethereum/evmc/pull/62

PR #334, opened 2018-08-19
Rename externalBalance to getExternalBalance

axic:
Counterpart of https://github.com/ewasm/design/pull/120

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -235,13 +235,13 @@ namespace hera {
       return Literal();
     }
 
-    if (import->base == Name("getBalance")) {
+    if (import->base == Name("getExternalBalance")) {
       heraAssert(arguments.size() == 2, string("Argument count mismatch in: ") + import->base.str);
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -235,13 +235,13 @@ namespace hera {
       return Literal();
     }
 
-    if (import->base == Name("getBalance")) {
+    if (import->base == Name("getExternalBalance")) {
```
jakelang: I would rather have things in line with existing naming schemes like RPC.
In that can, getExternal* would become get get*At which, in my opinion, makes more sense upon reading.

axic: This is in line with the EEI.

jakelang: Ok the name change should probably be introduced in a later revision, as it would probably break a lot of tests for the current one.

Maybe having multiple aliases for EEI methods would be good for backwards compatibility, at the cost of clarity

PR #336, opened 2018-08-19
Validate the argument range of getBlockHash

axic:


src/eei.cpp
```diff
@@ -261,6 +261,8 @@ namespace hera {
 
       HERA_DEBUG << "getBlockHash " << hex << number << " " << resultOffset << dec << "\n";
 
+      ensureCondition(number >= 0 && number <= 256, ArgumentOutOfRange, "getBlockHash only accepts a block number between 0 and 256");
```
chfast: That's not correct. It accepts range of `(current_block_number - 1, current_block_number - 257)` .

axic: Btw, we could also change `getblockhash` in EVMC to return an error code for reason of failure.

chfast: https://github.com/ethereum/evmc/issues/63

axic:
Actually this PR doesn't make sense. We have `getBlockHash` to return an error code, that's the way it should return if it is out of range.

The `ArgumentOutOfRange` exception makes more sense of invalid `gas` value supplied to `useGas` and `call.`

PR #346, opened 2018-08-21
Add more strict compiler flags for handling switch statements

axic:


src/CMakeLists.txt
```diff
@@ -3,6 +3,9 @@ find_package(Threads REQUIRED)
 set(hera_include_dir ${PROJECT_SOURCE_DIR}/include)
 get_filename_component(evmc_include_dir .. ABSOLUTE)
 
+# enforce some strictness around switch statements
+add_compile_options(-Wswitch-default -Wswitch-bool -Wimplicit-fallthrough)
```
chfast: 1. Instead of `-Wswitch-default` I prefer `-Wswitch` (already enabled) - it warns if not all enum cases are handled. Solidity is going this direction.
2. The `-Wswitch-bool` is enabled by default.
3. For `-Wimplicit-fallthrough` we must check compiler support. I could add this to Cable.

axic: Ture about `switch-bool` and `switch-default`.

I actually meant to have
> -Wswitch-enum
> Warn whenever a switch statement has an index of enumerated type and lacks a case for one or more of the named codes of that enumeration. case labels outside the enumeration range also provoke warnings when this option is used. The only difference between -Wswitch and this option is that this option gives a warning about an omitted enumeration code even if there is a default label.

But that seems to be gcc only (not present in clang).

axic: Created issue #373 to replace this PR.

PR #348, opened 2018-08-21
Split out debugging helpers into their own file (debugging.h)

axic:


axic:
> Everywhere you have HERA_DEBUG only the HERA_DEBUGGING check is not need.

Yeah, though I didn't wanted to change that in this PR in case some of those have side effects. Will fix that next.

PR #351, opened 2018-08-21
Consider most of the 256-bit values as bytes32 and keep difficulty as uint256

### LINKED ISSUE #318: Clarify endianness of memory access operations

axic:
Depends on #322.

Fixes #318.


axic:
This breaks evm2wasm, but I'm going to ignore that for the moment.

PR #355, opened 2018-08-22
Split Binaryen implementation and the core EthereumInterface

axic:
Part of #156.

Heavily based on #148.

### DIFF
--- a/src/eei.h
+++ b/src/eei.h
@@ -33,7 +33,7 @@ struct ExecutionResult {
   bool isRevert = false;
 };
 
-class EthereumInterface : public ShellExternalInterface {
+class EthereumInterface {
 public:
   explicit EthereumInterface(
     evmc_context* _context,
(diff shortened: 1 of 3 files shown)


src/eei.h
```diff
@@ -33,15 +33,15 @@ struct ExecutionResult {
   bool isRevert = false;
 };
 
-struct EthereumInterface : ShellExternalInterface {
+struct EthereumInterface {
```
chfast: Make it `class`.

src/eei.h
```diff
@@ -185,4 +175,33 @@ struct GasSchedule {
   static constexpr unsigned callNewAccount = 25000;
 };
 
+struct BinaryenEthereumInterface : ShellExternalInterface, public EthereumInterface {
```
chfast: Make it `class`.

src/eei.h
axic: Yeah not sure why it was still a struct.

PR #359, opened 2018-08-22
Move Binaryen execution code into its own file and class

axic:


src/hera.cpp
```diff
@@ -413,8 +315,8 @@ evmc_result hera_execute(
     }
 
     heraAssert(hera->wasm_engine == hera_wasm_engine::binaryen, "Unsupported wasm engine.");
-
-    ExecutionResult result = execute(context, run_code, state_code, *msg, meterInterfaceGas);
+    BinaryenEngine engine = BinaryenEngine{};
```
axic: Here when using `WasmEngine engine = BinaryenEngine{};` I had an error:
```
/Users/alex/Projects-Changed/hera/src/hera.cpp:326:16: error: variable type 'hera::WasmEngine' is an abstract class
    WasmEngine engine = BinaryenEngine{};
               ^
/Users/alex/Projects-Changed/hera/src/eei.h:38:27: note: unimplemented pure virtual method 'execute' in 'WasmEngine'
  virtual ExecutionResult execute(
                          ^
```

Suggestions to resolve?

chfast: You cannot do it this way. You need a reference or pointer.
It can be `WasmEngine& engine = BinaryenEngine{};` but this is not useful.

PR #366, opened 2018-08-26
Added mapping for engine name to enum

axic:


src/hera.cpp
```diff
@@ -409,17 +419,10 @@ int hera_set_option(
...
-#if HERA_WABT
-     if (strcmp(value, "wabt") == 0)
-       hera->wasm_engine = hera_wasm_engine::wabt;
-#endif
-#if HERA_WAVM
-     if (strcmp(value, "wavm") == 0)
-       hera->wasm_engine = hera_wasm_engine::wavm;
-#endif
-     return 1;
+    if (wasm_engine_names.count(name)) {
```
jakelang: I am pretty sure this should be `value`, not `name`.

src/hera.cpp
```diff
@@ -409,17 +419,10 @@ int hera_set_option(
...
-     if (strcmp(value, "wabt") == 0)
-       hera->wasm_engine = hera_wasm_engine::wabt;
-#endif
-#if HERA_WAVM
-     if (strcmp(value, "wavm") == 0)
-       hera->wasm_engine = hera_wasm_engine::wavm;
-#endif
-     return 1;
+     if (wasm_engine_names.count(name)) {
+       hera->wasm_engine = wasm_engine_names[name];
```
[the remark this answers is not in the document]
jakelang: I think the compiler infers it, as it also builds on linux clang. I still use the `string(name)` constructor to be explicit

src/hera.cpp
axic: Fixed.

PR #375, opened 2018-08-27
Run evm2wasm tests using runevm

axic:
Depends on #288, #513. Depends on releasing [runevm](https://github.com/axic/runevm).

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -166,14 +166,19 @@ defaults:
       command: |
         ~/build/evmc/test/evmc-vmtester ~/build/src/libhera.so
 
-#  evm2wasm-test: &evm2wasm-test
-#    run:
-#      name: "Run evm2wasm state tests"
-#      command: |
-#        if [[ $PRELOAD_ASAN ]]; then export LD_PRELOAD=/usr/lib/clang/8/lib/linux/libclang_rt.asan-x86_64.so; fi
-#        testeth --version
-#        testeth -t GeneralStateTests/stExample -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "add11" --evmc evm1mode=evm2wasm.cpp
-#        testeth -t GeneralStateTests/stStackTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --evmc evm1mode=evm2wasm.cpp
+  evm2wasm-test: &evm2wasm-test
+    run:
+      name: "Run evm2wasm state tests (using runevm contract)"
+      command: |
+        curl -o /tmp/runevm.wasm -L https://github.com/axic/runevm/releases/download/0.1.0-alpha2/runevm.wasm
+        test "$(shasum /tmp/runevm.wasm)" = "f9c11823f34bc2910bfa457342e3d6d00801ea5f  /tmp/runevm.wasm"
+        if [[ $PRELOAD_ASAN ]]; then export LD_PRELOAD=/usr/lib/clang/8/lib/linux/libclang_rt.asan-x86_64.so; fi
+        testeth --version
+        testeth -t GeneralStateTests/stExample -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --evmc evm1mode=runevm --evmc sys:runevm=/tmp/runevm.wasm
+        # This works, but takes too much time (4 minutes)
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -137,14 +137,24 @@ defaults:
...
         testeth -t GeneralStateTests/stStackTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --evmc evm1mode=evm2wasm.cpp
 
+  evm2wasm-contract-test: &evm2wasm-contract-test
+    run:
+      name: "Run evm2wasm state tests (using runevm contract)"
+      command: |
+        curl -o /tmp/runevm.wasm -L https://github.com/axic/runevm/releases/download/prototype/runevm.wasm
+        #test "$(shasum /tmp/runevm.wasm)" = "598f87f072f66e2269dd6919292b2934dbb20492  /tmp/runevm.wasm"
+        testeth --version
+        testeth -t GeneralStateTests/stExample -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "add11" --evmc evm1mode=evm2wasm --evmc sys:evm2wasm=/tmp/runevm.wasm
```
jakelang: I am pretty sure that you do not need to use `--evmc` preceding each option. Last I checked (possibly a previous version of EVMC), it worked fine with one `--evmc` invocation for all options.

axic: How would that work? `--evmc "evm1mode=evm2wasm sys:evm2wasm=/tmp/runevm.wasm"` ?

I do think it needs separate ones because that's how parsing is done by aleth.

cc @chfast

chfast: Checking...

chfast: Now you have to do `--evmc a=b --evmc x=y`, but I'm working on a change to allow `--evmc a=b x=y`.

axic: Does it make sense to have the latter supported?

chfast: What's the "latter"?

axic: `--evmc a=b x=y`

chfast: It's shorted isn't it?

axic: What happens if the value part of the option has spaces?

chfast: `--evmc "a=b and c" x=y`?

axic: I find it confusing when parameters can have spaces, e.g. if there are multiple EVMC options in one line I'd do `--evmc "a=b x=y"` and in taht case `b` or `y` cannot have spaces

chfast: Maybe, but I'm not going to write a custom option parser just for EVMC options. Now it is simple as finding first `=` and splitting the string accordingly.

If you want you can check whenever CLI11 library offers any fancy option grouping we could use.

axic: Well I guess the simple solution is not changing how it works currently? :)

chfast: It's not a change, it's an extension.

axic: Added in https://github.com/ethereum/aleth/pull/5234.

Okay, if the old way is still accepted then I have nothing against it.

axic:
I'm not sure for what case testeth prints this:
```
testeth 1.4.0rc7
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::program_options::validation_error> >'
  what():  the argument for option '--vm' is invalid
/bin/bash: line 3:  2144 Aborted                 testeth -t GeneralStateTests/stExample -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "add11" --evmc evm1mode=evm2wasm --evmc sys:evm2wasm=/tmp/runevm.wasm
Exited with code 134
```

axic:
@s1na it seems at least the `add11` test is still failing with runevm. Do you have a test case which works?

s1na:
@axic I think it might be due to https://github.com/wasmx/wasm-chisel/pull/85. The runevm [release](https://github.com/axic/runevm/releases/tag/0.1.0-alpha1) probably doesn't initialize memory. It passes for me locally (I build `runevm` and pass it through `chisel`'s master branch)

axic:
Ah ok, that might be it! Will push a new build of runevm to try.

Can you tell me which tests suites it passes?

s1na:
These should at least pass (possibly some others): `stExample`, `stStackTests`, `stShift` and `stCodeSizeLimit`

axic:
@s1na uploaded a version built with latest chisel but it still seems to be failing.

Added you to this repo so you can push to this PR, if you would like to play with it.

axic:
This is working now, by mistake the previous runevm binary didnt had the deployer code.

PR #377, opened 2018-08-27
Move WasmEngine to Hera instance

### LINKED ISSUE #372: Create proper abstraction so that we can use `WasmEngine` with multiple VMs
In https://github.com/ewasm/hera/blob/master/src/hera.cpp#L297 should be able to do something like:
```
WasmEngine engine;
switch (engine) {
  case wasm_engine::binaryen:
    engine = BinaryenEngine{};
    break;
  case wasm_engine::wabt:
    engine = WabtEngine{};
    break;
}
engine.execute(...);
```

chfast:
Keep the Wasm Engine by owned pointer in the Hera instance. Pointer allows future polymorphic behavior.

Fixes #372.

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -77,12 +77,24 @@ const map<string, hera_evm_mode> evm_mode_options {
 };
 
 struct hera_instance : evmc_instance {
-  hera_wasm_engine wasm_engine = hera_wasm_engine::binaryen;
+  std::unique_ptr<WasmEngine> engine{new BinaryenEngine};
   hera_evm_mode evm_mode = hera_evm_mode::reject;
   bool metering = false;
 
   hera_instance() noexcept : evmc_instance({EVMC_ABI_VERSION, "hera", hera_get_buildinfo()->project_version, nullptr, nullptr, nullptr, nullptr}) {}
 };
+
+int hera_create_wasm_engine(struct hera_instance *hera, hera_wasm_engine engine)
+{
+  switch (engine) {
+  case hera_wasm_engine::binaryen:
+    hera->engine.reset(new BinaryenEngine);
+    break;
+  default:
+    return 0;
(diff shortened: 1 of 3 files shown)


src/hera.cpp
```diff
@@ -405,8 +405,15 @@ int hera_set_option(
   }
 
   if (strcmp(name, "engine") == 0) {
-    if (wasm_engine_options.count(value)) {
-      hera->wasm_engine = wasm_engine_options.at(value);
+    auto it = wasm_engine_options.find(value);
```
chfast: I don't think the `wasm_engine_options` and `hera_wasm_engine` are needed any more. Here simple string switch will do.

axic: `string` switch? Is there such a thing in C++?

src/hera.cpp
```diff
@@ -405,8 +405,15 @@ int hera_set_option(
...
+    if (it != wasm_engine_options.end()) {
+      switch (it->second) {
+      case hera_wasm_engine::binaryen:
+        hera->engine.reset(new BinaryenEngine);
```
axic: If we're keeping the enum then I'd create a helper closer where the enum is defined: `hera_create_engine(hera_instance* hera, string const& name);`

axic:
Not fully sure we should disallow `WasmEngine` to have instance variables (with this change we are).  Will know better when wavm is implemented.

If we are, then `WasmEngine` should have a comment.

Are there any other options?

src/hera.cpp
axic: @chfast I made this change. Merging it this way, because I am not sure wabt/wavm will not need instance variables. Once we have a basic wabt/wavm code merged which works with this limitation I'd be happy to remove this extra indirection of having `enum hera_wasm_engine`.

PR #379, opened 2018-08-28
Use WasmEngine factory method to simplify the engine creation

chfast:
Part of #372.


chfast:
Don't ask me to do something and then do it yourself.

axic:
It's been 2 hours and no indication. Don't be salty.

PR #382, opened 2018-08-28
Revamp the README

axic:


chfast:
Why are you requesting a review and the merge it without any?

axic:
Because the readme needs more work anyway and can be done separately. I'd like to finish all my Hera work today.

PR #390, opened 2018-08-28
Update to EVMC 5.2.0

axic:


axic:
@chfast also added error mapping for outofrange, but not for the VM trap because we don't have it isolated between user trap and wasm trap.

PR #391, opened 2018-08-28
Split wasm engines into external/internal headers

axic:


axic:
@chfast there is probably a better way to do this. I'll let you take care.

axic:
> Move the contents of X-internal.h to X.cpp.

Isn't that against the main goal of having a clear separation of class definition (gives a good overview what the members/functions are) and implementation?

src/binaryen.cpp
```diff
@@ -25,6 +25,7 @@
 #include <wasm-validator.h>
 
 #include "binaryen.h"
+#include "binaryen-internal.h"
```
chfast: It's best when headers matching a .cpp file are included as first. Then you avoid situation when the headers misses any include.

axic:
@chfast did the split for wavm. Can you check if this is what you've meant?

src/binaryen.cpp
```diff
@@ -434,8 +467,10 @@ namespace hera {
     heraAssert(false, string("Unsupported import called: ") + import->module.str + "::" + import->base.str + " (" + to_string(arguments.size()) + "arguments)");
   }
 
+namespace {
+
 // NOTE: This should be caught during deployment time by the Sentinel.
-void BinaryenEngine::validate_contract(Module & module)
+void validate_contract(Module & module)
```
axic: @chfast Which is better, this in an anonymous namespace or an empty forward declaration of `wasm::Module` in `binaryen.h` - see the last commit for that.

axic:
> but this will mess up the other WAVM PR.

Shouldn't really affect it, once it is rebased. Currently it duplicates some of the commits already merged on a different PR.

src/binaryen.cpp
chfast: If possible, hide it from the header.

axic: It can stay in an anonymous namespace.

axic:
@chfast final version ^^

Issue #376, opened 2018-08-27
Fuzzing Hera


axic:
It should be fuzzed with random inputs to be executed (with and without the sentinel).

Most of the fuzzing corpus should be built with the successful test cases of ewasm tests suite. Goal is fuzzing the EEI implementation.

holiman:
Are we talking about consensus-testing, or just 'can I crash this thing?'. Basically, AFL-fuzz it like solidity/solfuzz or make something a bit more clever, like differential fuzzing against geth/parity ? 

Cc @guidovvranken

axic:
For now 
> can I crash this thing?

In 2-3 weeks (more realistically after devcon) we can move into consensus testing by running both Hera and Wagon. But first Wagon support needs to be finished in geth @gballet.

axic:
A quick and dirty way to crash on the desired exceptions is to add `abort()` into these three:
https://github.com/ewasm/hera/blob/master/src/hera.cpp#L371-L380

- `(InternalErrorException)`
- `(std::exception)`
- `(...)`

holiman:
Point me to a binary, build instructions, LMK what types of input it takes, and I'll get started :)

axic:
Use this branch: https://github.com/ewasm/hera/pull/392

Build with:
```
mkdir build
cd build
cmake -DBUILD_SHARED_LIBS=ON ..
make
```

Run with `aleth --vm ~/build/src/libhera.so --singlenet Byzantium <code goes here>`.

PR #392, opened 2018-08-28
Add fuzzing option to abort on internal errors

axic:
Part of #376.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ These are to be used via EVMC `set_option`:
 - `metering=true` will enable metering of bytecode at deployment using the [Sentinel system contract] (set to `false` by default)
 - `evm1mode=<evm1mode>` will select how EVM1 bytecode is handled
 - `sys:<alias/address>=file.wasm` will override the code executing at the specified address with code loaded from a filepath at runtime. This option supports aliases for system contracts as well, such that `sys:sentinel=file.wasm` and `sys:evm2wasm=file.wasm` are both valid. **This option is intended for debugging purposes.**
+- `fuzzing=true` will enable fuzzing mode, which aborts on an internal error (as opposed to transmitting the error code to the client). **This option is solely intended for fuzzing purposes.**
 
 ### evm1mode
 
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -88,6 +88,7 @@ struct hera_instance : evmc_instance {
 #error "No engine requested."
 #endif
   };
+  bool fuzzing = false;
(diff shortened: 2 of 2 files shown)


axic:
This is for afl.

README.md
```diff
@@ -44,6 +44,7 @@ These are to be used via EVMC `set_option`:
 - `metering=true` will enable metering of bytecode at deployment using the [Sentinel system contract] (set to `false` by default)
 - `evm1mode=<evm1mode>` will select how EVM1 bytecode is handled
 - `sys:<alias/address>=file.wasm` will override the code executing at the specified address with code loaded from a filepath at runtime. This option supports aliases for system contracts as well, such that `sys:sentinel=file.wasm` and `sys:evm2wasm=file.wasm` are both valid. **This option is intended for debugging purposes.**
+- `fuzzing=true` will enable fuzzing mode, which aborts on an internal error (as opposed to transmitting the error code to the client). **This option is solely intended for fuzzing purposes.**
```
chfast: How about naming it `abort-on-internal-error`?

src/hera.cpp
```diff
@@ -371,12 +372,18 @@ evmc_result hera_execute(
...
     ret.status_code = EVMC_INTERNAL_ERROR;
     HERA_DEBUG << "Unknown exception: " << e.what() << "\n";
+    if (hera->fuzzing)
+      abort();
   } catch (...) {
     ret.status_code = EVMC_INTERNAL_ERROR;
     HERA_DEBUG << "Totally unknown exception\n";
+    if (hera->fuzzing)
+      abort();
   }
```
chfast:
```suggestion
if (ret.status_code == EVMC_INTERNAL_ERROR)
  abort();
```

src/hera.cpp
```diff
@@ -371,12 +372,18 @@ evmc_result hera_execute(
   } catch (InternalErrorException const& e) {
     ret.status_code = EVMC_INTERNAL_ERROR;
     HERA_DEBUG << "InternalError: " << e.what() << "\n";
+    if (hera->fuzzing)
```
chfast: You can replace these checks with single one suggested below.

src/hera.cpp
axic: True. Though probably I prefer this way because the branch is only triggered in case of an error.

README.md
axic: Any shorter suggestion?

chfast: No.

axic: A suggestion: "errormode", with possible value "abort-on-internal-error".

axic:
This may be obsolete now.

PR #393, opened 2018-08-30
Query evmc_tx_context once in each execution

axic:
Also independently discovered by @jakelang.

### DIFF
--- a/src/eei.h
+++ b/src/eei.h
@@ -67,6 +67,9 @@ public:
     // set sane defaults
     m_result.returnValue = std::vector<uint8_t>{};
     m_result.isRevert = false;
+
+    // cache the transaction context here
+    m_context->fn_table->get_tx_context(&m_tx_context, m_context);
   }
 
 protected:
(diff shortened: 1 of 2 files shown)


src/eei.h
```diff
@@ -67,6 +67,9 @@ class EthereumInterface {
     // set sane defaults
     m_result.returnValue = std::vector<uint8_t>{};
     m_result.isRevert = false;
+
+    // cache the transaction context here
+    m_context->fn_table->get_tx_context(&m_tx_context, m_context);
```
jakelang: perhaps lazily populate this (upon first call by the EEI) and store it as a `pair<bool, tx_context>` that gets checked on each EEI call using it?

It would be a bit of a micro-optimization but contracts that do not need the tx_context don't need to call EVMC for it.

axic: But then again at every single EEI method querying that would have the extra branch checking that. I think it is better to assume some of these will be neeed and that simplifies code a lot.

@jakelang @chfast I would even consider changing EVMC to pass that down with `execute`.

axic: Created issue: https://github.com/ethereum/evmc/issues/120

PR #394, opened 2018-08-30
Initial support for WAVM as an engine

### LINKED ISSUE #158: Support wavm as a VM
Depends on #156.

poemm:
Early version of wavm execution in Hera. Should compile with `cmake -DHERA_WAVM=ON`. May need some work.

Closes #158.

Depends on #401.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -143,6 +143,15 @@ defaults:
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getGasLeftUseAllGas" --evmc engine=wabt
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "returnPredefinedData" --evmc engine=wabt
 
+  test-wavm: &test-wavm
+    run:
+      name: "Test shared Hera (wavm)"
+      command: |
+        if [[ $PRELOAD_ASAN ]]; then export LD_PRELOAD=/usr/lib/clang/6.0/lib/linux/libclang_rt.asan-x86_64.so; fi
+        testeth --version
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "useGas" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "createFromTransaction" --evmc engine=wavm
+
   evmc-test: &evmc-test
     run:
       name: "Run evmc tests"
@@ -190,6 +199,7 @@ jobs:
       - *fetch-tests
       - *test
       - *test-wabt
+      - *test-wavm
       - *evmc-test
       - *evm2wasm-test
 
--- a/circle.yml
+++ b/circle.yml
@@ -152,9 +152,11 @@ defaults:
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "useGas" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "createFromTransaction" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "storageStore" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "storageLoad" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getCallDataSize" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "callDataCopy" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "callDataCopy256" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getGasLeft" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getGasLeftUseAllGas" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "returnPredefinedData" --evmc engine=wavm
 
--- a/circle.yml
+++ b/circle.yml
@@ -151,6 +151,12 @@ defaults:
         testeth --version
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "useGas" --evmc engine=wavm
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "createFromTransaction" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "storageStore" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getCallDataSize" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "callDataCopy" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "callDataCopy256" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "getGasLeftUseAllGas" --evmc engine=wavm
+        testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "returnPredefinedData" --evmc engine=wavm
 
   evmc-test: &evmc-test
     run:
--- a/src/eei.h
+++ b/src/eei.h
@@ -72,7 +72,12 @@ public:
     m_context->fn_table->get_tx_context(&m_tx_context, m_context);
   }
 
+// WAVM host functions access this interface through an instance,
+// which requires public methods.
+// TODO: update upstream WAVM to have a context (user data) passed down.
+#if HERA_WAVM == 0
 protected:
+#endif
   virtual size_t memorySize() const = 0 ;
   virtual void memorySet(size_t offset, uint8_t value) = 0;
   virtual uint8_t memoryGet(size_t offset) = 0;
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -126,9 +126,11 @@ namespace wavm_host_module {
       IR::ObjectType type,
       Runtime::Object*& outObject) override
     {
+      outObject = nullptr;
       auto namedInstance = moduleNameToInstanceMap.get(moduleName);
-      outObject = Runtime::getInstanceExport(*namedInstance, exportName);
-      return true;
+      if (namedInstance)
+          outObject = Runtime::getInstanceExport(*namedInstance, exportName);
+      return outObject != nullptr;
     }
   };
 } // namespace wavm_host_module
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -142,6 +142,15 @@ ExecutionResult WavmEngine::execute(
 ) {
   HERA_DEBUG << "Executing with wavm...\n";
 
+  // Collect garbage left by previous runs.
+  //
+  // GCPointer<> marks objects to be deleted upon moving out of scope, however
+  // since all the objects are part of this call, `collectGarbage` at the end
+  // won't actually clean anything.
+  //
+  // Here we try to clean up after previous runs.
+  Runtime::collectGarbage();
+
   // set up a new ethereum interface just for this contract invocation
   ExecutionResult result;
   WavmEthereumInterface interface{context, state_code, msg, result, meterInterfaceGas};
@@ -200,7 +209,8 @@ ExecutionResult WavmEngine::execute(
 
   // clean up
   wavm_host_module::interface.pop();
-  // TODO: does this actually clean up all GCPointers above?
+
+  // This likely won't clear any memory, but we can be hopeful.
   Runtime::collectGarbage();
 
   return result;
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -72,9 +72,21 @@ namespace wavm_host_module {
   }
 
 
+  DEFINE_INTRINSIC_FUNCTION(ethereum, "getGasLeft", U64, getGasLeft)
+  {
+    return static_cast<U64>(interface.top()->eeiGetGasLeft());
+  }
+
+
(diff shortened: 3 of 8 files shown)


src/eei.h
```diff
@@ -69,7 +69,9 @@ class EthereumInterface {
     m_result.isRevert = false;
   }
 
+#if HERA_WAVM==OFF
```
jakelang: why is this conditionally compiled for WAVM only?

I think these methods should be protected anyway?

jakelang: Ok, I think it would then make sense to leave a note to remove it once the WAVM EEI implementation is more developed and actually calls those methods.

poemm: Updated it with a comment.

My vote is to remove `protected:` entirely.

src/wavm.cpp
```diff
@@ -0,0 +1,263 @@
...
+  */
+  DEFINE_INTRINSIC_FUNCTION(ethereum, "revert", void, revert, U32 dataOffset, U32 length)
+  {
+    printf("revert  dataOffset=%i  length=%i\n",dataOffset,length);
+    try{
+      interface->eeiRevert(dataOffset, length);
+    }
+    catch(hera::HeraException e){
+      //gracefully return from the wavm invokation
+      throwException(Runtime::Exception::calledAbortType);
```
axic: We need a way to carry over the hera exception type.

src/wavm.cpp
```diff
@@ -0,0 +1,263 @@
...
+    for (int i=0;i<length; i++){
+      U8 ui = i;
+      code.push_back(ui);
+    }
+    // put code into the memory
+    for (int i=0;i<length; i++)
+      buffer[i] = code[i];
+    */
+
+    interface->eeiCallDataCopy(resultOffset, dataOffset, length);
```
axic: Every single eei method can throw exceptions.

axic:
Regarding the namespace vs. class: is there a way to pass user data down the resolver to the functions?

That is the same way `wabt` works, they are still static functions, but receive a user data pointer in which we can pass down the context.

src/wavm.cpp
poemm: Yes, good point. I will look for a way to carry over the hera exception type.

src/wavm.cpp
poemm: "Early version .... May need some work."

poemm:
I tried to mirror the architecture of wabt in hera. Unfortunately, I have not found an easy way to pass arbitrary data to a wavm host function. I will look again.

src/wavm.cpp
```diff
@@ -0,0 +1,263 @@
...
+std::unique_ptr<WasmEngine> WavmEngine::create()
+{
+  return std::unique_ptr<WasmEngine>{new WavmEngine};
+}
+
+
+namespace wavm_host_module {
+
+  // first a pointer to the ethereum interface, which is used in host functions
+  WavmEthereumInterface* interface;
```
axic: This cannot stay like this. The `WavmEthereumInterface` doesn't only hold helpers for the host functions, but the call context they are in.

Having this as a global means any ewasm bytecode doing a call to another contract wouldn't work.

axic:
Need to add CI support (search for wabt in `circle.yml`)

src/wavm.cpp
poemm: I see. Wavm and Hera architectures seem incompatible.

axic: Just open an issue upstream asking for a context to be passed down.

poemm: They will point me to the existing way to pass something to host functions, but it is complicated with mutexes, garbage collectors, custom create functions instead of constructors. I will try this but might take time and might be unsuccessful.

src/wavm.cpp
```diff
@@ -158,27 +84,26 @@ namespace wavm_host_module {
...
+    printf("call  gas=%lu  addressOffset=%i  valueOffset=%i  dataOffset=%i  dataLength=%i\n", gas, addressOffset, valueOffset, dataOffset, dataLength);
 
-    try{
-      interface->eeiUseGas(gas);
-    }
-    catch(hera::HeraException e){
-      throwException(Runtime::Exception::calledAbortType);
-    }
+    // save a pointer to the interface before the call
+    WavmEthereumInterface* interface_before_call = interface;
```
axic: I think instead of this `interface` should be a `std::stack` where the context is pushed before call, popped after and used with `interface.top()` in each EEI host function.

axic: Well actually it should not be pushed/popped in call, but in `wavm_execute`, the entry point to this.

src/wavm.h
```diff
@@ -34,7 +34,7 @@ class WavmEthereumInterface : public EthereumInterface {
 
 private:
   // These assume that m_wasmMemory was set prior to execution.
-  size_t memorySize() const override { return 0; } //TODO
+  size_t memorySize() const override { return m_wasmMemory->numPages * 65536; } //TODO 65536 bytes = 64 kb; not sure if this works
```
axic: This is used to check for overflow when reading/writing to the linear memory. As long as `memoryGet(memorySize())` is valid, this should be ok.

src/wavm.h
```diff
@@ -0,0 +1,66 @@
...
+    EthereumInterface(_context, _code, _msg, _result, _meterGas)
+  {}
+
+  void setWasmMemory(Runtime::MemoryInstance* _wasmMemory) {
+    m_wasmMemory = _wasmMemory;
+  }
+
+private:
+  // These assume that m_wasmMemory was set prior to execution.
+  size_t memorySize() const override { return m_wasmMemory->numPages * 65536; } //TODO 65536 bytes = 64 kb; not sure if this works
```
axic: This doesn't work now:
```
In file included from /home/builder/project/src/hera.cpp:35:
/home/builder/project/src/wavm.h:37:59: error: member access into incomplete type 'Runtime::MemoryInstance'
  size_t memorySize() const override { return m_wasmMemory->numPages * 65536; } //TODO 65536 bytes = 64 kb; not sure if this works
                                                          ^
deps/src/wavm/Include/Runtime/Runtime.h:76:50: note: forward declaration of 'Runtime::MemoryInstance'
        DECLARE_OBJECT_TYPE(ObjectKind::memory, Memory, MemoryInstance);
                                                        ^
```

src/wavm.cpp
poemm: You are correct. `interface` is a global defined just below `namespace wavm_host_module {`. Yes, the old pointer is stored during execution of `eeiCall()` so it can be restored to how it was before the call.

chfast: Oh, if you back it up, replace and restore the backup in the end that will work too. I don't see the full picture, but I feel more comfortable with putting such magic features in execute() rather than in eeiCall().

poemm:
Committed the requested updates. Now need to test things like exception handling and recursive calls. THANK YOU for all of the useful tips.

src/wavm.cpp
```diff
@@ -179,7 +171,10 @@ ExecutionResult WavmEngine::execute(
     }
   );
 
-  //delete wavm_host_module::interface; //TODO: clean up, not if there is anybody else using it
+  // clean up
+  WavmEthereumInterface* interface_popped = wavm_host_module::interface.top();
+  wavm_host_module::interface.pop();
+  //delete interface_popped; //TODO how to clean up?
```
axic: You could use `std::unique_ptr` to deal with that (or is it `shared_ptr`? @chfast can advise).

circle.yml
```diff
@@ -131,6 +131,14 @@ defaults:
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "useGas" --evmc engine=wabt
         testeth -t GeneralStateTests/stEWASMTests -- --testpath tests --vm ~/build/src/libhera.so --singlenet Byzantium --singletest "createFromTransaction" --evmc engine=wabt
 
+  test-wavm: &test-wavm
```
axic: This needs to be added into the `steps` part of `linux-clang-shared-release`, otherwise it is not executed.

src/wavm.cpp
```diff
@@ -0,0 +1,183 @@
...
+      Runtime::invokeFunctionChecked(wavm_context, functionInstance, invokeArgs);
+    },
+    [&](Runtime::Exception&& exception) {
+      //TODO: handle wavm exception here
+    }
+  );
+
+  // clean up
+  WavmEthereumInterface* interface_popped = wavm_host_module::interface.top();
+  wavm_host_module::interface.pop();
```
axic: @chfast should execute (because of push/pop) be used `lock_guard` (https://en.cppreference.com/w/cpp/thread/lock_guard)?

src/wavm.cpp
```diff
@@ -0,0 +1,183 @@
...
+  HashMap<std::string, Runtime::Object*> extraEthereumExports; //empty for current ewasm stuff
+  Runtime::GCPointer<Runtime::ModuleInstance> ethereumHostModule;
+  ethereumHostModule = Intrinsics::instantiateModule(compartment, wavm_host_module::INTRINSIC_MODULE_REF(ethereum), "ethereum", extraEthereumExports);
+  // prepare contract module to resolve links against host module
+  wavm_host_module::MyResolver myResolver(compartment);
+  myResolver.moduleNameToInstanceMap.set("ethereum", ethereumHostModule);
+  Runtime::LinkResult linkResult = linkModule(moduleAST, myResolver);
+
+  // instantiate contract module
+  Runtime::ModuleInstance* moduleInstance = instantiateModule( compartment, moduleAST, std::move(linkResult.resolvedImports), "arbitrary_name_of_contract_module");
```
axic: Can you really move `resolvedImports` here?

axic: What is `"arbitrary_name_of_contract_module"` exactly for?

poemm: I will retest `resolvedImports`, don't remember why it is done like this, but it worked in early tests.

`"arbitrary_name_of_contract_module"` is the name of the module, arbitrary for now. Some people prefer things like this to document themselves, but I will change it to the empty string.

circle.yml
poemm: Yes, will add it to steps.

axic:
The README should also be updated:
- change "in progress" to "limited support" under the wavm heading
- add wavm as an option on the line talking about the `engine` option

src/wavm.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+  // get memory for easy access in host functions
+  // wavm_host_module::interface->memory = asMemory(getInstanceExport(moduleInstance, "memory"));
+  wavm_host_module::interface.top()->setWasmMemory( asMemory(getInstanceExport(moduleInstance, "memory")) );
+
+  // invoke the main function
+  Runtime::FunctionInstance* functionInstance = asFunctionNullable(getInstanceExport(moduleInstance, "main"));
```
axic: There should be a `heraAssert(functionInstance, "main function not found");` here.

src/wavm.cpp
```diff
@@ -0,0 +1,183 @@
...
+  wavm_host_module::interface.push(new WavmEthereumInterface{context, state_code, msg, result, meterInterfaceGas});
+
+  // first parse module
+  std::vector<U8> codeBytes = static_cast< std::vector<U8> > (code);
+  IR::Module moduleAST; 
+  loadBinaryModule(codeBytes.data(), codeBytes.size(), moduleAST);
+
+  // next set up the host module. Note: in ewasm, we create a new VM for each call to a module, so we must instantiate a new host module for each of these VMs, this is inefficient, but OK for prototyping.
+  // compartment is like the Wasm store, represents the VM, has lists of globals, memories, tables, and also has wavm's runtime stuff
+  Runtime::Compartment* compartment = Runtime::createCompartment();
```
axic: These are pointers:
- do we need safeguards for null pointers here?
- who manages the lifecycle of these pointers?

axic: It seems we need to take care of the lifetime of this one:

```
Compartment* Runtime::createCompartment() { return new Compartment(); }
```

src/wavm.cpp
```diff
@@ -0,0 +1,183 @@
...
+  // context stores the compartment and some other stuff
+  Runtime::Context* wavm_context = Runtime::createContext(compartment);
```
axic: Same here, wavm don't seem to manage this pointer. However it loosely references `compartment`...

src/wavm.cpp
axic: Lifecycle of `moduleInstance` seems not to be managed by wavm.

src/wavm.cpp
axic: This is just a pointer into `moduleInstance`, safe to "ignore", but should have a comment explaining it.

poemm:
Made most requested changes. Now passing all tests for eei methods `useGas`, `getCallDataSize`, `callDataCopy`, `revert`, `finish`. Ready for final review and merge.

axic:
I think we should merge #401 first so that ASan tests this too.

src/wavm.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  // first parse module
+  std::vector<U8> codeBytes = static_cast< std::vector<U8> > (code);
+  IR::Module moduleAST; 
+  bool loadedSuccess = loadBinaryModule(codeBytes.data(), codeBytes.size(), moduleAST);
+  heraAssert(loadedSuccess, "wavm couldn't parse module into syntax tree");
+
+  // next set up the host module. Note: in ewasm, we create a new VM for each call to a module, so we must instantiate a new host module for each of these VMs, this is inefficient, but OK for prototyping.
+  // compartment is like the Wasm store, represents the VM, has lists of globals, memories, tables, and also has wavm's runtime stuff
+  Runtime::GCPointer<Runtime::Compartment> compartment = Runtime::createCompartment();
```
axic: Not fully convinced we should use `Runtime::GCPointer` and not `std::unique_ptr` if because of that we need to call `Runtime::collectGarbage()`. However if we need to call that anyway, then I am fine.

src/wavm.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+  // invoke the main function
+  Runtime::GCPointer<Runtime::FunctionInstance> functionInstance = asFunctionNullable(getInstanceExport(moduleInstance, "main"));
+  heraAssert(functionInstance, "wavm couldn't find main function");
+  std::vector<IR::Value> invokeArgs; //empty for current ewasm stuff
+  Runtime::catchRuntimeExceptions(  //this is how WAVM's try/catch for exceptions
+    [&] {
+      try { //this is how Hera's try/catch for exceptions
+        Runtime::invokeFunctionChecked(wavm_context, functionInstance, invokeArgs);
+      } catch (EndExecution const&) {
```
axic: What about the other exceptions? Are they passed through?

src/wavm.cpp
poemm: WAVM uses some custom abstracted constructor-like things, and I don't think that they can be cleaned up through `std::unique_ptr`.

axic: @chfast can you take a look?

src/wavm.cpp
poemm: Both WAVM and Hera exceptions seem to be handled properly for all test cases for opcodes `useGas`, `getCallDataSize`, `callDataCopy`, `revert`, `finish`.

Try it yourself, all of the test cases are `useGas`, `getCallDataSize`, `callDataCopy`, `callDataCopy256`, `revert`, `revertVeryLong`, `returnPredefinedData`, and `returnVeryLong`.

src/wavm.cpp
chfast: When you use it and it does not crash you are good.

axic:
> Author: user <36397285+poemm@users.noreply.github.com>

@poemm your git user is like the above. Would it be possible to include a name or a nick in place of `user`? At least `poemm <poemm@users.noreply.github.com>` would be better for future license tracking.

If so, you can update that in gitconfig with (and create a new dummy commit after, I can squash stuff into it):
```
$ git config set user.email poemm@users.noreply.github.com
$ git config set user.name poemm
```

(Though an actual name would be preferred.)

axic:
Tried moving off `GCPointer` to `unique_ptr`, but unfortunately a lot of the objects are only forward declared and `unique_ptr` doesn't work well with that.

axic:
Apparently can't call in the destructor:
```
*** No errors detected
Assertion failed at /home/builder/build/deps/src/wavm/Lib/Platform/POSIX.cpp(867): !pthread_mutex_lock((pthread_mutex_t*)&pthreadMutex)
Call stack:
  host!/home/builder/build/src/libhera.so!Platform::Mutex::lock()+22
  host!/home/builder/build/src/libhera.so!Runtime::collectGarbage()+46
  host!/home/builder/build/src/libhera.so!hera::WavmEngine::~WavmEngine()+44
  host!/home/builder/build/src/libhera.so!<unknown>
```

axic:
Added all the tests from wabt and it failed:
```
unknown location(0): fatal error: in "GeneralStateTests/stEWASMTests": memory access violation at address: 0x00000008: no mapping at fault address
```

Turns out `storageLoad` is not implemented, just `storageStore`, however this means that loading any wasm with an unsupported import will result in a segfault.

@poemm any idea where to fix that?

poemm:
Great work! You gave three possibilities, another possibility follows. Redefine `execute()` to
`execute(){ run(); collectGarbage(); }` where the body of `run()` is the current body of `execute()`. This way all `GCPointer`s will go out of scope before calling `collectGarbage()`. What do you think?

PR #396, opened 2018-09-04
Implement storageStore/storageLoad/getGasLeft/getCallValue in wabt

### LINKED ISSUE #380: Implement storageStore/storageLoad/getGasLeft/getCallValue in wabt
The wabt backend should be extended to support:
- `storageStore`
- `storageLoad`
- `getGasLeft`
- `getCallValue`

See how the current methods are implemented in `wabt.cpp` and see the definition of the above in `eei.h`. Also worth looking into `binaryen.cpp` how it does the wrapping for these.

Additionally circleci must be updated to run some tests which cover these.

This is an easy/medium complexity task.

bakaoh:
Fix #380


axic:
Well done, thank you!

PR #408, opened 2018-09-08
CMake: Compiler flags

### LINKED ISSUE #373: CMake support "-Wimplicit-fallthrough" and "-Wswitch-enum"
These are compiler specific so need additional code to turn them on when possible.

From https://github.com/ewasm/hera/pull/346#discussion_r212819782

chfast:
Closes #373 

This adds `-Wswitch-enum` flag for Hera library.
The `-Wimplicit-fallthrough` was added earlier with `cable_configure_compiler()`.

Also it updates the Cable to avoid the "old version" warning.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,8 +31,10 @@ set(PROJECT_VERSION 0.1.0)
 cable_set_build_type(DEFAULT RelWithDebInfo CONFIGURATION_TYPES Debug;Release;RelWithDebInfo)
 
 cable_configure_compiler()
-# TODO: fix the warnings instead
-add_compile_options(-Wno-pedantic)
+if(CABLE_COMPILER_GNULIKE)
+    # TODO: fix the warnings instead
+    add_compile_options(-Wno-pedantic)
+endif()
 
 include(ProjectBinaryen)
 


src/eei.cpp
```diff
@@ -530,6 +530,23 @@ namespace hera {
...
+      case EVMC_CALL_DEPTH_EXCEEDED:
+      case EVMC_STATIC_MODE_VIOLATION:
+      case EVMC_PRECOMPILE_FAILURE:
+      case EVMC_CONTRACT_VALIDATION_FAILURE:
+      case EVMC_ARGUMENT_OUT_OF_RANGE:
+      case EVMC_WASM_UNREACHABLE_INSTRUCTION:
+      case EVMC_WASM_TRAP:
+      case EVMC_INTERNAL_ERROR:
+      case EVMC_REJECTED:
       default:
```
axic: If you have all the cases, should we keep `default`?

After seeing this, not convinced this is the best way. What do you think?

chfast: The `default` is needed on older compiler, otherwise then cannot prove that the function always returns. Besides you want to handle unknown internal errors.

I'm not sure, this is very explicit. Probably a bit overkill.

chfast: The `default` is needed on older compiler, otherwise then cannot prove that the function always returns. Besides you want to handle unknown internal errors.

I'm not sure, this is very explicit. Probably a bit overkill.

chfast:
Removed the `-Wswitch-enum` commit.

PR #420, opened 2018-09-17
Get more info about WAVM memory leaks

chfast:
If you build WAVM with RelWithDebInfo the ASan will report source locations for memory leaks.


cmake/ProjectWAVM.cmake
```diff
@@ -36,7 +36,7 @@ ExternalProject_Add(wavm
     -DCMAKE_CXX_COMPILER=${CMAKE_CXX_COMPILER}
     -DCMAKE_C_COMPILER=${CMAKE_C_COMPILER}
     -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
-    -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
axic: Should we just use this by default?

PR #425, opened 2018-09-18
Run returnVeryLong/revertVeryLong on CI for wavm

axic:
Depends on #496.


axic:
@poemm the returndata code actually fails with:
```
getReturnDataSize
Unknown exception: bad_function_call
/home/builder/project/test/tools/libtesteth/ImportTest.cpp(753): error: in "GeneralStateTests/stEWASMTests": returnVeryLong on Byzantium: Expected another postState hash! expected: 0x18dd6b7c1ba4192101c8d1d9705065c853b90eecc9bd1f534ec0bdbff3e40df3 actual: 0xf90f3f4e474d95fdc166d5cc54f81d31a3f9c9bf0d55e877fab210b436e33af1 in Byzantium data: 0 gas: 0 val: 0
```

axic:
Basically `catchRuntimeException` doesn't let `HeraException` to pass through. We need to wrap it into a wavm exception and then unwrap it.

PR #428, opened 2018-09-19
Move depth and sender balance checks earlier in eeiCall

axic:
Follow up of #435.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -494,22 +494,20 @@ namespace hera {
       // Start with base call gas
       takeInterfaceGas(GasSchedule::call);
 
+      if (m_msg.depth >= 1024)
+        return 1;
+
       // These checks are in EIP150 but not in the YellowPaper
       // Charge valuetransfer gas if value is being transferred.
       if ((kind == EEICallKind::Call || kind == EEICallKind::CallCode) && !isZeroUint128(call_message.value)) {
         takeInterfaceGas(GasSchedule::valuetransfer);
 
-        // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
-        if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
-          takeInterfaceGas(GasSchedule::callNewAccount);
-      }
-
-      if (m_msg.depth >= 1024)
-        return 1;
-
-      if ((kind == EEICallKind::Call) || (kind == EEICallKind::CallCode)) {
         if (!enoughSenderBalanceFor(call_message.value))
           return 1;
+
+        // Only charge callNewAccount gas if the account is new and non-zero value is being transferred per EIP161.
+        if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
+          takeInterfaceGas(GasSchedule::callNewAccount);
       }
 
       // This is the gas we are forwarding to the callee.
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -496,10 +496,11 @@ namespace hera {
 
       // These checks are in EIP150 but not in the YellowPaper
       // Charge valuetransfer gas if value is being transferred.
-      // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
       if ((kind == EEICallKind::Call || kind == EEICallKind::CallCode) && !isZeroUint128(call_message.value)) {
         takeInterfaceGas(GasSchedule::valuetransfer);
-        if (!m_context->fn_table->account_exists(m_context, &call_message.destination))
+
+        // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
+        if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
           takeInterfaceGas(GasSchedule::callNewAccount);
       }
 
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -497,9 +497,9 @@ namespace hera {
       // These checks are in EIP150 but not in the YellowPaper
       // Charge valuetransfer gas if value is being transferred.
       // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
-      if (!isZeroUint128(call_message.value)) {
+      if ((kind == EEICallKind::Call || kind == EEICallKind::CallCode) && !isZeroUint128(call_message.value)) {
         takeInterfaceGas(GasSchedule::valuetransfer);
-        if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
+        if (!m_context->fn_table->account_exists(m_context, &call_message.destination))
           takeInterfaceGas(GasSchedule::callNewAccount);
       }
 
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -461,6 +461,13 @@ namespace hera {
         dataLength << dec << "\n";
 #endif
 
+      if ((kind == EEICallKind::Call) || (kind == EEICallKind::CallCode)) {
```
axic: @lrettig remember why this was only added for call/callcode?

Apparently the logic came from aleth, cc @chfast

chfast: Only CALL and CALLCODE have value argument.
The depth check should be done for all calls.

axic: > Only CALL and CALLCODE have value argument.

Ok so that part is an optimisation.

axic: > The depth check should be done for all calls.

Fixed that in #429.

axic:
Tests fail because gas usage is different. In the previous code the `extra_gas` was not refunded if there was a value or depth issue.

@chfast apparently aleth supposed to do it this way (but I'm not sure). Any comments?

chfast:
> We have an inconsistency in that case. Create doesn't charge the base fee if one of those two conditions are hit.

So the CREATE is free then?

axic:
> So the CREATE is free then?

At least in Hera in those cases 😉

lrettig:
I fixed the gas logic in https://github.com/ewasm/hera/pull/255 to bring it inline with aleth and trinity, and to get state tests passing. I'm concerned that this might break that but haven't tried the tests against this change yet.

axic:
@chfast @jakelang @lrettig can you review again?

src/eei.cpp
```diff
@@ -492,18 +492,23 @@ namespace hera {
...
       // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
       if (!isZeroUint128(call_message.value)) {
-        extra_gas += GasSchedule::valuetransfer;
+        takeInterfaceGas(GasSchedule::valuetransfer);
         if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
-          extra_gas += GasSchedule::callNewAccount;
+          takeInterfaceGas(GasSchedule::callNewAccount);
       }
 
-      // This check is in EIP150 but not in the YellowPaper
```
lrettig: This comment `// This check is in EIP150 but not in the YellowPaper` appears to have disappeared, but I find it helpful. Could you add it back above the two previous `takeInterfaceGas` calls?

src/eei.cpp
```diff
@@ -492,18 +492,23 @@ namespace hera {
       evmc_result call_result;
 
       // Start with base call gas
-      int64_t extra_gas = GasSchedule::call;
+      takeInterfaceGas(GasSchedule::call);
 
       // Charge valuetransfer gas if value is being transferred.
       // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
       if (!isZeroUint128(call_message.value)) {
```
lrettig:
```suggestion
      // These checks are in EIP150 but not in the YellowPaper
      if (!isZeroUint128(call_message.value)) {
```

src/eei.cpp
```diff
@@ -492,18 +492,23 @@ namespace hera {
...
-        extra_gas += GasSchedule::valuetransfer;
+        takeInterfaceGas(GasSchedule::valuetransfer);
         if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
```
jakelang: uhhhhh there are other calls with value transfer that must be covered

axic:
@cdetrio please also check.

Once we agree on this, we need to refill the tests.

src/eei.cpp
```diff
@@ -492,18 +492,25 @@ namespace hera {
...
       // Start with base call gas
-      int64_t extra_gas = GasSchedule::call;
+      takeInterfaceGas(GasSchedule::call);
 
+      // These checks are in EIP150 but not in the YellowPaper
       // Charge valuetransfer gas if value is being transferred.
-      // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
-      if (!isZeroUint128(call_message.value)) {
-        extra_gas += GasSchedule::valuetransfer;
+      if ((kind == EEICallKind::Call || kind == EEICallKind::CallCode) && !isZeroUint128(call_message.value)) {
```
lrettig: For `CALL`, with `call_message.value == 0`, where the account doesn't already exist, you're not charging `GasSchedule::callNewAccount` here. aleth checks `zeroValueTransferChargesNewAccountGas` which has to do with EIP158 (cf. https://github.com/ethereum/aleth/blob/37ead28874c5adb380d4c60eadff161ca9736ece/libethcore/EVMSchedule.h#L86). Can you confirm that this is intentional?

src/eei.cpp
jakelang: is this gonna be fixed? @axic

src/eei.cpp
```diff
@@ -492,18 +492,25 @@ namespace hera {
...
+
+        // Only charge callNewAccount gas if the account is new and value is being transferred per EIP161.
         if ((kind == EEICallKind::Call) && !m_context->fn_table->account_exists(m_context, &call_message.destination))
-          extra_gas += GasSchedule::callNewAccount;
+          takeInterfaceGas(GasSchedule::callNewAccount);
       }
 
-      // This check is in EIP150 but not in the YellowPaper
-      takeInterfaceGas(extra_gas);
+      if (m_msg.depth >= 1024)
```
chfast: This is weird (although correct) that the depth is checked _after_ e.g. new account creation cost is already charged.

src/eei.cpp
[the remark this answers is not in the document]
chfast: This suggestion looks useful.

axic: That aleth code is super confusing and I have no idea what each of those random rules mean and on what hard fork they are enabled. We are copying byzantium behaviour. More or less.

src/eei.cpp
gballet: And in EVM geth the depth is checked before the account is charged; in its current state, wagon relies a lot on the default code and it would misbehave and force me to write a lot of redundant code.

src/eei.cpp
```diff
@@ -492,18 +492,25 @@ namespace hera {
...
+        return 1;
+
+      if ((kind == EEICallKind::Call) || (kind == EEICallKind::CallCode)) {
+        if (!enoughSenderBalanceFor(call_message.value))
```
gballet: There's an issue here with wagon, thanks for helping me identify it ;)

src/eei.cpp
chfast: Well, I'm confused every time I try to understand how calls works. 

First of all, the order does not matter. All those checks are applied to the caller so if any of these fails the caller execution will be aborted.

I'm for moving the depth check up front because it's the cheapest one.
My ideal order would be:

1. depth check (no state access)
2. value transfer additional cost check (no state access)
3. balance check (current account access)
4. account creation additional cost check (external account lookup)


Side note: the order will matter if we decide to always user REVERT semantic (i.e. dropping OOG behavior).

src/eei.cpp
chfast: This is ok. This is byzantium behavior.

axic:
@gballet are you happy with the latest revision ?

gballet:
> @gballet are you happy with the latest revision ?

as happy as a Frenchman can be

PR #432, opened 2018-09-28
disable exports check until metering contract is fixed

cdetrio:
this should be used on the testnet until the metering contract is fixed.


axic:
Please just use this branch on ansible, but don't merge it as it won't be removed properly (also won't pass the tests when merged).

axic:
Closing this since the latest sentinel in https://github.com/ewasm/sentinel-rs works without this as confirmed by @jwasinger

PR #433, opened 2018-10-03
Rethrow EEI exceptions in WAVM

poemm:
Passes the following tests `useGas`, `getCallDataSize`, `callDataCopy`, `callDataCopy256`, `revert`, `returnPredefinedData`, `revertVeryLong`, and `returnVeryLong`.

Each host function which can throw a hera exception has a try-catch to catch and record this exception, trigger a WAVM trap so that it can clean up. Once WAVM returns, we re-throw the original hera exception.

Don't know if this is what you wanted. If not, I will close.

### DIFF
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -344,27 +344,26 @@ ExecutionResult WavmEngine::execute(
   bool meterInterfaceGas
 ) {
 
-  // clear stuff from previous run
-  wavm_host_module::eei_exception_ptr = nullptr;
-
-  // hope and prayer
-  Runtime::collectGarbage();
-
   // execute the contract
   ExecutionResult result;
   try {
+    // Unneeded safety measure.
+    wavm_host_module::eei_exception_ptr = nullptr;
+
     result = internalExecute(context, code, state_code, msg, meterInterfaceGas);
 
     // re-throw exception if there was one
-    if (wavm_host_module::eei_exception_ptr)
+    if (wavm_host_module::eei_exception_ptr) {
       std::rethrow_exception (wavm_host_module::eei_exception_ptr);
+      wavm_host_module::eei_exception_ptr = nullptr;
+    }
   } catch (EndExecution const&) {
       HERA_DEBUG << "caught Hera's EndExecution\n";
       // This exception is ignored here because we consider it to be a success.
       // It is only a clutch for POSIX style exit()
   }
 
-  // clean up this run, this is done here after leaving the scope of internalExecute()
+  // And clean up mess left by this run.
   Runtime::collectGarbage();
 
   return result;
@@ -437,28 +436,19 @@ ExecutionResult WavmEngine::internalExecute(
     [&] {
         vector<IR::Value> invokeArgs;
         Runtime::invokeFunctionChecked(wavm_context, mainFunction, invokeArgs);
-        // wavm may have cleaned up the memory, so make sure we can't access it
-        wavm_host_module::interface.top()->setWasmMemory(nullptr);
     },
     [&](Runtime::Exception&& exception) {
       HERA_DEBUG << "caught WAVM's Runtime::Exception\n";
-      // wavm may have cleaned up the memory, so make sure we can't access it
-      wavm_host_module::interface.top()->setWasmMemory(nullptr);
-      // if WAVM threw an exception and we did not, then make a note so we can throw a corresponding hera exception
+
+      // Throw a proper EEI exception from this.
       if (wavm_host_module::eei_exception_ptr == nullptr)
-        wavm_host_module::eei_exception_ptr = hera::VMTrap{Runtime::describeException(exception)};
+        wavm_host_module::eei_exception_ptr = make_exception_ptr{hera::VMTrap{Runtime::describeException(exception)}};
     }
   );
 
   // clean up
   wavm_host_module::interface.top()->setWasmMemory(nullptr);
   wavm_host_module::interface.pop();
-  compartment = nullptr;
-  wavm_context = nullptr;
-  ethereumHostModule = nullptr;
-  moduleInstance = nullptr;
-  mainFunction = nullptr;
-  Runtime::collectGarbage();
 
   return result;
 }
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -77,7 +77,6 @@ namespace wavm_host_module {
 
   // set up for re-throwing exceptions, this is not clean, but Hera and WAVM are incompatible without some awkwardness
   std::exception_ptr eei_exception_ptr = nullptr;
-  bool VMTrapFlag = false;
 
   // the host module is called 'ethereum'
   DEFINE_INTRINSIC_MODULE(ethereum)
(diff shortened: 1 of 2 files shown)


src/wavm.cpp
```diff
@@ -193,9 +227,49 @@ ExecutionResult WavmEngine::execute(
...
+
+  // execute the contract
   ExecutionResult result = internalExecute(context, code, state_code, msg, meterInterfaceGas);
-  // And clean up mess left by this run.
+
+  // clean up this run, this is done here after leaving the scope of internalExecute()
   Runtime::collectGarbage();
+
+  // re-throw exception if there was one
+  if (wavm_host_module::exceptionIdx != wavm_host_module::HeraExceptionsEnum::None){
```
jakelang: maybe simplify using a helper?

src/wavm.cpp
```diff
@@ -193,9 +227,49 @@ ExecutionResult WavmEngine::execute(
   evmc_message const& msg,
   bool meterInterfaceGas
 ) {
+
+  // clear stuff from previous run
+  wavm_host_module::exceptionIdx = wavm_host_module::HeraExceptionsEnum::None;
+  wavm_host_module::exceptionString = "";
+  // hope and prayer
+  Runtime::collectGarbage();
```
jakelang: why do we collect garbage redundantly here? (other call at L241)

src/wavm.cpp
```diff
@@ -143,13 +157,33 @@ namespace wavm_host_module {
...
   }
 
 
   DEFINE_INTRINSIC_FUNCTION(ethereum, "revert", void, revert, U32 dataOffset, U32 length)
   {
-    interface.top()->eeiRevert(dataOffset, length);
+    try {
+      interface.top()->eeiRevert(dataOffset, length);
+    } catch (InvalidMemoryAccess const& e) {
+      HERA_DEBUG<<"caught Hera's InvalidMemoryAccess\n";
```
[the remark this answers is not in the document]
poemm: I agree that this debug message could be more descriptive. How about outputting the original exception message `HERA_DEBUG << e.what()`? Or `HERA_DEBUG << "caught Hera's exception: "<< e.what() << ". Will exit WAVM invokation now." `? Or just removing this debug message altogether?

src/wavm.cpp
poemm: Yes, this is a "hope and prayer" that it will clean up something. Removing this is OK if you insist.

src/wavm.cpp
poemm: Could do it if you insist. But `hera_execute()` does a similar long `switch-case` inside of an `if` without a helper function.

src/wavm.cpp
```diff
@@ -143,13 +157,33 @@ namespace wavm_host_module {
 
   DEFINE_INTRINSIC_FUNCTION(ethereum, "finish", void, finish, U32 dataOffset, U32 length)
   {
-    interface.top()->eeiFinish(dataOffset, length);
```
axic: This also throws `EndExecution` which isn't a failure.

src/wavm.cpp
```diff
@@ -143,13 +157,33 @@ namespace wavm_host_module {
...
+      throwException(Runtime::Exception::calledUnimplementedIntrinsicType); //or maybe reachedUnreachableType
+    }
   }
 
 
   DEFINE_INTRINSIC_FUNCTION(ethereum, "revert", void, revert, U32 dataOffset, U32 length)
   {
-    interface.top()->eeiRevert(dataOffset, length);
+    try {
+      interface.top()->eeiRevert(dataOffset, length);
```
axic: This also throws `EndExecution` with revert set.

axic:
All the EEI methods can throw multiple exceptions (potentially 7 different ones). Given we implement ~10 EEI methods in this file and each exception catching is around 10 lines of code if we do not generalise this then this will take ~700 lines of code.

poemm:
Thank you for the tip, `std::current_exception` would be more concise and maintainable.

For this to work, we need a `std::exception_ptr` in the scope of the host functions, I see nowhere else to declare this ptr but next to `stack<WavmEthereumInterface*>`. This ptr must be nulled after we rethrow it (so that we don't accidentally reuse this exception in different calls), I see nowhere else to null it but at the bottom of each `call*` host function, and also at the beginning of `WavmEngine::execute()`. Not sure if there will be any ambiguities or edge cases.

poemm:
Granted @chfast 's request. All tests in the original post are still passing.

axic:
> testeth 1.4.0rc7

This PR is using a very old Hera. It needs to be rebased to make tests pass. Also there are some typos the linter is picking up.

poemm:
I fetched upstream, then did `git rebase upstream/master wavm_rethrow_exceptions`, then fixed typos, then force pushed.

src/wavm.cpp
```diff
@@ -71,10 +77,16 @@ unique_ptr<WasmEngine> WavmEngine::create()
...
 
+
 namespace wavm_host_module {
+
   // first the ethereum interface(s), the top of the stack is used in host functions
   stack<WavmEthereumInterface*> interface;
 
+  // set up for re-throwing exceptions, this is not clean, but Hera and WAVM are incompatible without some awkwardness 
+  std::exception_ptr my_exception_ptr;
+  bool VMTrapFlag;
```
chfast: Init to `false`.

src/wavm.cpp
```diff
@@ -57,6 +59,10 @@ class WavmEthereumInterface : public EthereumInterface {
     m_wasmMemory = _wasmMemory;
   }
 
+  void nullWasmMemory() {
```
axic: Instead of this you could use `setWasmMemory(nullptr)`.

src/wavm.cpp
```diff
@@ -120,6 +136,9 @@ namespace wavm_host_module {
   DEFINE_INTRINSIC_FUNCTION(ethereum, "storageStore", void, storageStore, U32 pathOffset, U32 valueOffset)
   {
     interface.top()->eeiStorageStore(pathOffset, valueOffset);
+    // print stuff for debugging
+    //interface.top()->debugPrintMem(1, valueOffset, 32);
+    //interface.top()->debugPrintStorage(1, pathOffset);
```
axic: Need to be removed before merging.

poemm:
@chfast I upgraded wavm to the version used in https://github.com/ewasm/hera/pull/422 and implemented exception rethrow in all host functions. The tests listed above pass, but there is still a memory leak. Can we upgrade to the newest version of wavm to see if the memory leak remains? I tried upgrading, but there were build errors -- the cmake stuff may be less difficult for you.

axic:
WAVM has a lot of commits lately, including [this documentation on exceptions](https://github.com/WAVM/WAVM/commit/c1d2ba945edcc9d87e9fc0044777ab1850a1c2a6).

Maybe exception handled is better now?

axic:
I think this actually isn't needed. The problem in #425 wasn't due to the fact that exceptions are not passed through, but that the `interface` stack was not freed if an exception occurred and thus in a case of a contract calling another after returning from a call it will messed up.

PR #435, opened 2018-10-04
Charge interface gas early (part 1)

axic:
This fixes everything, but `call` and `storageStore`.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -138,19 +138,19 @@ namespace hera {
   {
       HERA_DEBUG << "getAddress " << hex << resultOffset << dec << "\n";
 
-      storeAddress(m_msg.destination, resultOffset);
-
       takeInterfaceGas(GasSchedule::base);
+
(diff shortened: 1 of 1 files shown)


src/eei.cpp
```diff
@@ -618,11 +627,12 @@ namespace hera {
 
       ensureCondition(!(m_msg.flags & EVMC_STATIC), StaticModeViolation, "selfDestruct");
 
+      takeInterfaceGas(GasSchedule::selfdestruct);
```
axic: This should come before the ensureCondition.

axic:
> Explanation?

The discussion in #428. Basically these things are free if the conditions are not met without this change.

axic:
@hugo-dc @lrettig tests need to be refilled with this.

hugo-dc:
@axic refilled tests in this PR: https://github.com/ewasm/tests/pull/97

gballet:
I assume that the tests are failing because they haven't been updated

PR #436, opened 2018-10-04
Reject contract if start function is defined (binaryen)

axic:


src/binaryen.cpp
```diff
@@ -503,6 +503,12 @@ ExecutionResult BinaryenEngine::execute(
   );
 
   // NOTE: Most of this should be caught during deployment time by the Sentinel.
+  ensureCondition(
+    !module.start.is(),
```
axic: This may need to be `isNull()`

axic:
@hugo-dc this should be working, can you test it with the tests PR?

PR #438, opened 2018-10-05
Run verifyContract on the bytecode returning during deployment

axic:
Depends on https://github.com/ewasm/tests/pull/91


jakelang:
Why is `createFromTransaction` failing?

I am guessing because the checks in the EEI version of `verifyContract` are not implemented and the cases for invalid contracts succeed wrongly.

PR #445, opened 2018-10-23
Update bumpversion config, set version to 0.1.0-alpha.0

chfast:
Same rules as https://github.com/ethereum/aleth/issues/5276


axic:
Is this following the schema from Aleth?

chfast:
Yes.

PR #449, opened 2018-10-26
Remove code duplication of loadModule in binaryen

axic:


src/binaryen.h
```diff
@@ -41,6 +41,10 @@ class BinaryenEngine : public WasmEngine {
 
 private:
   void verifyContract(wasm::Module & module);
+
+  /// Parses and loads a Wasm module.
+  /// Don't ask, Module has no copy constructor, hence the reference.
+  void loadModule(std::vector<uint8_t> const& code, wasm::Module & module);
```
jakelang: Perhaps we could move this to a virtual method so that any abstract VM should have to implement it?

axic: It would be super abstract 😉

src/binaryen.cpp
```diff
@@ -517,10 +509,8 @@ ExecutionResult BinaryenEngine::execute(
   return result;
 }
 
-void BinaryenEngine::verifyContract(vector<uint8_t> const& code)
+void BinaryenEngine::loadModule(vector<uint8_t> const& code, wasm::Module & module)
```
chfast: Can't you return it?

axic: See the comment in the header:
> /// Don't ask, Module has no copy constructor, hence the reference.

PR #451, opened 2018-10-26
Ensure that Wasm preamble is valdiated properly

axic:


axic:
@chfast updated

axic:
This code would be significantly simpler if we'd remove evm2wasm and the fallback entirely.

PR #458, opened 2018-11-24
Revert on calls from addresses with 128bit+ balances

### LINKED ISSUE #456: Hera's gas accounting in callSenderBalanceExceeds128Bits
The `callSenderBalanceExceeds128Bits` test poses some problem with wagon and as a DDoS attack vector.

The initial state is as such:
```yaml
  pre:
    a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
      balance: 100000000000
      code: ''
      nonce: '0x4'
      storage: {}
    deadbeef00000000000000000000000000000001:
      balance: 100000000000
      code: ''
      nonce: ''
      storage: {}
    deadbeef00000000000000000000000000000000:
      balance: 340282366920938463463374607431768211600
      nonce: ''
      storage: {}
```

and the expected output is:

```wast
  expect:
    - indexes:
        data: !!int -1
        gas: !!int -1
        value: !!int -1
      network:
        - ALL
      result:
        # Balances remain the same as because of:
        # InternalError: Account balance (or transaction value) exceeds 128 bits
        a94f5374fce5edbc8e2a8697c15331677e6ebf0b:
          balance: 100000000000
        deadbeef00000000000000000000000000000000:
          balance: 340282366920938463463374607431768211600
          storage: {
            # Nothing should be store due to InternalError
            1: "0"
          }
        deadbeef00000000000000000000000000000001:
          balance: 100000000000
          storage: {}
```

This happens because upon [checking the sender's account balance for the value transfer](https://github.com/ewasm/hera/blob/v0.2.0/src/eei.cpp#L496), Hera also checks that [the account balance itself is lower than 128 bits](https://github.com/ewasm/hera/blob/v0.2.0/src/eei.cpp#L808) and if it is more, it will interrupt the program execution and not charge the sender's account anything.

I propose that this behavior be changed, at least to charge the transaction gas (21000), for the following reasons:

  * Because the wagon implementation relies on a lot of the original geth code, not charging the transaction gas cost would require a significant rewrite.
  * In the unlikely event that someone has a balance of more than 340282366920938463463374607431768211456, well, good for them. Why should we care?
  * In that case, this person could craft a huge number of transactions and DDoS the network without even being charged for it.

So I suggest to replace the 128 bit account balance check with a 256 bit one.

gballet:
Fixes #456

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -781,7 +781,7 @@ namespace hera {
 
   void EthereumInterface::storeUint128(evmc_uint256be const& src, uint32_t dstOffset)
   {
-    heraAssert(!exceedsUint128(src), "Account balance (or transaction value) exceeds 128 bits.");
+    ensureCondition(!exceedsUint128(src), ArgumentOutOfRange, "Account balance (or transaction value) exceeds 128 bits.");
     storeMemoryReverse(src.bytes + 16, dstOffset, 16);
   }
 
@@ -810,7 +810,7 @@ namespace hera {
 
   unsigned __int128 EthereumInterface::safeLoadUint128(evmc_uint256be const& value)
   {
-    heraAssert(!exceedsUint128(value), "Account balance (or transaction value) exceeds 128 bits.");
+    ensureCondition(!exceedsUint128(value), ArgumentOutOfRange, "Account balance (or transaction value) exceeds 128 bits.");
     unsigned __int128 ret = 0;
     for (unsigned i = 16; i < 32; i++) {
       ret <<= 8;


gballet:
@axic @chfast apparently I don't have the rights to request a review on this repo

gballet:
FYI CI falls because https://github.com/ewasm/tests/pull/105 hasn't been merged

gballet:
looking forward to that comment :innocent:

axic:
The way to fix is by changing the `heraAssert` in to `ensureCondition` (with `OutOfRange`) in `storeUint128` and `safeLoadUint128`.

src/eei.h
```diff
@@ -162,7 +162,7 @@ class EthereumInterface {
   /* Checks for overflow and safely charges gas for variable length data copies */
   void safeChargeDataCopy(uint32_t length, unsigned baseCost);
 
-  bool enoughSenderBalanceFor(evmc_uint256be const& value) const;
+  bool enoughSenderBalanceFor(evmc_uint256be const& value);
```
axic: Shouldn't remove `const` here IIRC.

PR #454, opened 2018-11-11
Update to wabt 1.0.6

axic:
1.0.6 gives us two benefits (amongst others):
- https://github.com/WebAssembly/wabt/pull/886: fixes host trapping
- https://github.com/WebAssembly/wabt/pull/882: speeds up the build process

PR #467, opened 2018-12-13
Remove redundant usage of wabt:: namespace

axic:
Split off #454.


axic:
@chfast though to be honest I've preferred the explicit namespace, because there are same fairly generic names in there.

PR #472, opened 2018-12-20
wabt rethrow exception in finish

poemm:
Exceptions in Hera would not allow Wabt invocation to free anything, resulting in a memory leak. This pull request catches and records Hera exceptions, uses https://github.com/WebAssembly/wabt/pull/886 to allow Wabt to clean up, then rethrows the Hera exception.

Implemented for the `finish` eei method, still needs to be implemented for the rest.

### DIFF
--- a/src/wabt.cpp
+++ b/src/wabt.cpp
@@ -49,19 +49,23 @@ public:
     ExecutionResult & _result,
     bool _meterGas
   ):
-    EthereumInterface(_context, _code, _msg, _result, _meterGas)
-  {}
+    EthereumInterface(_context, _code, _msg, _result, _meterGas){
+      exception_ptr_for_rethrowing = nullptr;
+    }
 
   // TODO: improve this design...
   void setWasmMemory(interp::Memory* _wasmMemory) {
     m_wasmMemory = _wasmMemory;
   }
 
+  exception_ptr exception_ptr_for_rethrowing;
+
 private:
   // These assume that m_wasmMemory was set prior to execution.
   size_t memorySize() const override { return m_wasmMemory->data.size(); }
   void memorySet(size_t offset, uint8_t value) override { m_wasmMemory->data[offset] = static_cast<char>(value); }
   uint8_t memoryGet(size_t offset) override { return static_cast<uint8_t>(m_wasmMemory->data[offset]); }
+
 
   interp::Memory* m_wasmMemory;
 };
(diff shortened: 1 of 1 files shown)


src/wabt.cpp
```diff
@@ -57,12 +57,15 @@ class WabtEthereumInterface : public EthereumInterface {
     m_wasmMemory = _wasmMemory;
   }
 
+  std::exception_ptr my_exception_ptr;
```
axic: Can you please not go with `my_exception_ptr`, rather something like `last_exception_ptr` ?

src/wabt.cpp
```diff
@@ -104,7 +107,15 @@ ExecutionResult WabtEngine::execute(
     ) {
       int64_t gas = static_cast<int64_t>(args[0].value.i64);
       // FIXME: handle host trap here
-      interface.eeiUseGas(gas);
+      try {
+        interface.eeiUseGas(gas);
+      } catch (HeraException const& e) {
```
axic: There's no need to name these arguments.

src/wabt.cpp
axic: No need to prefix with `std::`

src/wabt.cpp
```diff
@@ -260,14 +333,22 @@ ExecutionResult WabtEngine::execute(
   // FIXME: really bad design
   interface.setWasmMemory(env.GetMemory(0));
 
+  // for rethrowing exception
+  interface.my_exception_ptr = nullptr;
```
axic: I don't think this needs to be null'd here.

axic:
Did anybody check if wabt isn't already rethrowing non-wabt exceptions?

src/wabt.cpp
poemm: Ok.

src/wabt.cpp
poemm: Ok.

src/wabt.cpp
poemm: Ok.

src/wabt.cpp
poemm: Ok.

poemm:
> Did anybody check if wabt isn't already rethrowing non-wabt exceptions?

They don't.

poemm:
@chfast Please confirm that there are no detected memory leaks.

poemm:
(Responding to https://github.com/ewasm/hera/pull/473#issuecomment-451942837 .)

> Let's re-review it. Maybe there is something valuable there.

@chfast I re-reviewed this and don't see anything else interesting here. I think that this PR was useful because it showed that there is a way to solve memory leaks, which give you a hint for your better solution.

chfast:
>  I re-reviewed this and don't see anything else interesting here. I think that this PR was useful because it showed that there is a way to solve memory leaks, which give you a hint for your better solution.

Yes, I'm for closing this one.

PR #473, opened 2019-01-07
Fix WABT exception passing

chfast:
Before the WABT was built without exception support so the stack was not properly cleaned up.


poemm:
Once this is merged, I will close rethrowing exceptions https://github.com/ewasm/hera/pull/472 .

chfast:
> Once this is merged, I will close rethrowing exceptions #472 .

Let's re-review it. Maybe there is something valuable there.

PR #476, opened 2019-01-17
Upgrade EVMC to 6.0.2

chfast:


axic:
We can marge this, but has zero effect on Hera.

We really need to release EVMC 6.1.0 and support it here.

PR #484, opened 2019-01-31
Use bytes / bytes_view to handle buffers

chfast:


axic:
Can you split the CI stuff into a new PR?

axic:
Do you see any speed difference? Or is it just code clarity at the expense of needing C++17?

chfast:
No speed was measured.

chfast:
No speed was measured.

src/eei.h
```diff
@@ -53,13 +58,13 @@ class EthereumInterface {
 public:
   explicit EthereumInterface(
     evmc_context* _context,
-    std::vector<uint8_t> const& _code,
+    std::vector<uint8_t> const& _code,  // TODO: Change arg type to bytes or bytes_view.
```
axic: Changing this requires a lot of work or what is the reason postponing it?

src/binaryen.cpp
```diff
@@ -509,10 +509,12 @@ ExecutionResult BinaryenEngine::execute(
...
 }
 
-void BinaryenEngine::loadModule(vector<uint8_t> const& code, wasm::Module & module)
+void BinaryenEngine::loadModule(bytes_view code, wasm::Module& module)
 {
   try {
-    wasm::WasmBinaryBuilder parser(module, reinterpret_cast<vector<char> const&>(code), false);
+    auto const b = reinterpret_cast<char const*>(code.begin());
+    auto const e = reinterpret_cast<char const*>(code.end());
+    wasm::WasmBinaryBuilder parser(module, {b, e}, false);
```
axic: These three lines are way less clear to me compared to the original one.

src/eei.h
```diff
@@ -24,6 +26,9 @@
 
 namespace hera {
 
+using bytes = std::basic_string<uint8_t>;
+using bytes_view = std::basic_string_view<uint8_t>;
```
axic: Moving these to `helpers.h` may enable sharing them across the code.

src/binaryen.cpp
chfast: At least vector is not casted.

chfast:
> The benefit here is mostly readability?

No. Compatibility, i.e. you can easily convert any `uint8_t` buffer (vector, C array, string, etc) to `bytes_view`.

The build with libc++ is failing because it does not like `basic_fstream<uint8_t>`. I invented this on the fly, have to check if that's compliant with C++.

src/hera.cpp
```diff
@@ -131,19 +131,19 @@ pair<evmc_status_code, vector<uint8_t>> callSystemContract(
   if (result.release)
     result.release(&result);
 
-  return make_pair(result.status_code, ret);
+  return {result.status_code, ret};
```
axic: You want to merge such simplifications separately?

src/hera.cpp
```diff
@@ -273,12 +273,12 @@ evmc_result hera_execute(
     heraAssert(hera->engine, "Wasm engine not set.");
     WasmEngine& engine = *hera->engine;
 
-    ExecutionResult result = engine.execute(context, run_code, state_code, *msg, meterInterfaceGas);
+    ExecutionResult result = engine.execute(context, {run_code.data(), run_code.size()}, state_code, *msg, meterInterfaceGas);
```
axic: Is there no nicer way to do this? Cannot one pass `run_code` directly?

src/eei.cpp
```diff
@@ -246,8 +245,9 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
       safeChargeDataCopy(length, GasSchedule::extcode);
 
       evmc_address address = loadAddress(addressOffset);
-      // TODO: optimise this so no vector needs to be created
-      vector<uint8_t> codeBuffer(length);
+      // TODO: optimise this so no copy needs to be created
+      bytes codeBuffer;
+      codeBuffer.resize(length);
```
axic: I see below you use `data(length, 0)` to zero-fill. Can you do the same here?

src/binaryen.cpp
```diff
@@ -509,10 +509,13 @@ ExecutionResult BinaryenEngine::execute(
...
 }
 
-void BinaryenEngine::loadModule(vector<uint8_t> const& code, wasm::Module & module)
+void BinaryenEngine::loadModule(bytes_view code, wasm::Module& module)
 {
   try {
-    wasm::WasmBinaryBuilder parser(module, reinterpret_cast<vector<char> const&>(code), false);
+    auto const b = reinterpret_cast<char const*>(code.begin());
+    auto const e = reinterpret_cast<char const*>(code.end());
+    std::vector<char> codeCopy{b, e};
```
axic: Is there no way to avoid the copy here just because change the type form `uint8_t` to `char`?

chfast: No other than improving Binaryen API.

axic:
Ah, this actually passes tests just codecov is complaining.

src/eei.h
```diff
@@ -201,9 +206,9 @@ class EthereumInterface {
 
   evmc_tx_context m_tx_context{};
   evmc_context* m_context = nullptr;
-  std::vector<uint8_t> const& m_code;
+  bytes m_code;
```
axic: This should be `bytes_view`.

src/binaryen.cpp
axic: Improving how? Binaryen already takes a reference.

src/eei.cpp
```diff
@@ -267,8 +266,8 @@ void WasmEngine::collectBenchmarkingData()
       safeChargeDataCopy(length, GasSchedule::extcode);
 
       evmc_address address = loadAddress(addressOffset);
-      // TODO: optimise this so no vector needs to be created
-      vector<uint8_t> codeBuffer(length);
+      // TODO: optimise this so no copy needs to be created
+      bytes codeBuffer(length, 0);
```
axic: What is the `0` argument?

src/eei.h
```diff
@@ -83,13 +86,13 @@ class EthereumInterface {
...
-    std::vector<uint8_t> const& _code,
+    bytes_view _code,
     evmc_message const& _msg,
-    ExecutionResult & _result,
+    ExecutionResult& _result,
     bool _meterGas
   ):
     m_context(_context),
-    m_code(_code),
+    m_code{_code.begin(), _code.end()},
```
axic: Is this a copy here?

axic:
@chfast collapsed the outdated comments, but after a review there are still unanswered questions. Can you please take a look?

src/binaryen.cpp
chfast: The problem is Binaryen forces you to pass a vector.

PR #486, opened 2019-02-04
Benchmarking instrumentation

### LINKED ISSUE #485: -DHERA_BENCHMARKING=ON
Can we add flag `-DHERA_BENCHMARKING=ON` when compiling Hera, and include the following code in each engine's interface to print benchmarking data? @chfast 

```
#if HERA_BENCHMARKING
#include <fstream>      // std::fstream
#include <time.h>       // clock_t, clock
#endif

...

#if HERA_BENCHMARKING
clock_t t = clock();
#endif

...parse/validate/compile...

#if HERA_BENCHMARKING
t = clock() - t;
std::fstream fs;
fs.open ("runtime_data_wavm_compile.csv", std::fstream::out | std::fstream::app);
fs << ", " << ((float)t)/CLOCKS_PER_SEC;
fs.close();
#endif

...

#if HERA_BENCHMARKING
clock_t t2 = clock();
#endif

...invoke main...

#if HERA_BENCHMARKING
t2 = clock() - t2;
std::fstream fs2;
fs2.open ("runtime_data_wavm_invoke.csv", std::fstream::out | std::fstream::app);
fs2 << ", " << ((float)t2)/CLOCKS_PER_SEC;
fs2.close();
#endif
```

chfast:
Closes #485.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -15,10 +15,11 @@
  */
 
 #include <array>
-#include <vector>
-#include <sstream>
-#include <iostream>
+#include <fstream>
 #include <iomanip>
+#include <iostream>
+#include <sstream>
+#include <vector>
 
 #include "debugging.h"
 #include "eei.h"
@@ -59,6 +60,9 @@ void WasmEngine::collectBenchmarkingData()
 
   std::cerr << "Time [us]: " << to_us(instantiationDuration + executionDuration) << " = "
             << to_us(instantiationDuration) << " + " << to_us(executionDuration) << "\n";
+
+  std::ofstream f{"hera_benchmarks.csv", std::ios::out | std::ios::app};
+  f << to_us(instantiationDuration) << ',' << to_us(executionDuration) << '\n';
 }
 
 #if HERA_DEBUGGING
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -50,19 +50,20 @@ bool WasmEngine::benchmarkingEnabled = false;
 
 void WasmEngine::collectBenchmarkingData()
 {
-  constexpr auto to_us = [](clock::duration d) {
-    return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
+  // Convert duration to string with microsecond units.
+  constexpr auto to_us_str = [](clock::duration d) {
+    return std::to_string(std::chrono::duration_cast<std::chrono::microseconds>(d).count());
   };
 
   const auto now = clock::now();
   const auto instantiationDuration = executionStartTime - instantiationStartTime;
   const auto executionDuration = now - executionStartTime;
 
-  std::cerr << "Time [us]: " << to_us(instantiationDuration + executionDuration) << " = "
-            << to_us(instantiationDuration) << " + " << to_us(executionDuration) << "\n";
-
-  std::ofstream f{"hera_benchmarks.csv", std::ios::out | std::ios::app};
-  f << to_us(instantiationDuration) << ',' << to_us(executionDuration) << '\n';
+  const auto log = "Time [us]: " + to_us_str(instantiationDuration + executionDuration) +
+                   " (instantiation: " + to_us_str(instantiationDuration) +
+                   ", execution: " + to_us_str(executionDuration) + ")\n";
+  std::cerr << log;
+  std::ofstream{"hera_benchmarks.log", std::ios::out | std::ios::app} << log;
 }
 
 #if HERA_DEBUGGING
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -45,6 +45,22 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
 }
 }  // namespace
 
+bool WasmEngine::benchmarkingEnabled = false;
+
+void WasmEngine::collectBenchmarkingData()
+{
+  constexpr auto to_us = [](clock::duration d) {
+    return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
+  };
+
+  const auto now = clock::now();
+  const auto instantiationDuration = executionStartTime - instantiationStartTime;
+  const auto executionDuration = now - executionStartTime;
+
+  std::cerr << "Time [us]: " << to_us(instantiationDuration + executionDuration) << " = "
+            << to_us(instantiationDuration) << " + " << to_us(executionDuration) << "\n";
+}
+
 #if HERA_DEBUGGING
   void EthereumInterface::debugPrintMem(bool useHex, uint32_t offset, uint32_t length)
   {
--- a/src/eei.h
+++ b/src/eei.h
@@ -16,6 +16,7 @@
 
 #pragma once
 
+#include <chrono>
 #include <vector>
 
 #include <evmc/evmc.h>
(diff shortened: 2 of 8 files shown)


src/eei.h
```diff
@@ -47,6 +50,42 @@ class WasmEngine {
...
+
+  void executionFinished() noexcept
+  {
+    if (benchmarkingEnabled)
+    {
+      constexpr auto to_us = [](clock::duration d) {
+        return std::chrono::duration_cast<std::chrono::microseconds>(d).count();
+      };
+
+      auto executionDuration = clock::now() - executionStartTime;
```
poemm: `executionDuration` is not interesting to me other than to compute invocation time as `executionDuration - instantiationDuration`. But `instantiationDuration` is interesting.

src/hera.cpp
```diff
@@ -398,6 +398,14 @@ evmc_set_option_result hera_set_option(
     return EVMC_SET_OPTION_SUCCESS;
   }
 
+  if (strcmp(name, "benchmark") == 0) {
```
poemm: So we would run testeth with command line argument `benchmarking=true`?

poemm:
Thank you for this. Looks good. I see no problems. @axic do you insist on preprocessor flags?

axic:
> @axic do you insist on preprocessor flags?

Does it matter?

poemm:
> Does it matter?

Hmm. Depends on who you ask. This PR is OK for me. But I respect your design goals, so if you want preprocessor flags, then we will use preprocessor flags. I am blocked on this so I hope that you give guidance soon.

src/hera.cpp
chfast: Yes, by `--evmc benchmark=true`.

poemm:
Gah, benchmarking already has scripts to parse the previous stderr statements, my fault for not telling you. Thank you for improving the stderr statements, it may be helpful to include the words "instantiation" and "execution" to make it understandable, but after that please leave the stderr statements so that I only have to write my parsing script once. Also, the file may be useful, but please make it print the same thing as stderr.

BTW this design overwrites `instantiationStartTime` when a contract calls another contract. Which is OK for benchmarking since benchmarks are one contract calling the contract to be benchmarked.

src/eei.cpp
```diff
@@ -45,6 +46,25 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
 }
 }  // namespace
 
+bool WasmEngine::benchmarkingEnabled = false;
```
axic: I think this should go into the header.

src/eei.cpp
```diff
@@ -45,6 +46,25 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
...
+  };
+
+  const auto now = clock::now();
+  const auto instantiationDuration = executionStartTime - instantiationStartTime;
+  const auto executionDuration = now - executionStartTime;
+
+  std::cerr << "Time [us]: " << to_us(instantiationDuration + executionDuration) << " = "
+            << to_us(instantiationDuration) << " + " << to_us(executionDuration) << "\n";
+
+  std::ofstream f{"hera_benchmarks.csv", std::ios::out | std::ios::app};
```
axic: If we're creating files I think that should be documented in the readme.

axic:
> BTW this design overwrites `instantiationStartTime` when a contract calls another contract. Which is OK for benchmarking since benchmarks are one contract calling the contract to be benchmarked.

I don't think so. Each execution is a new instance of the engine. The only problem is in the output the children come first.

src/eei.cpp
chfast: That would define the variable multiple times.

axic: Why?

chfast: Because the header is included in multiple compilation units. You will not be able to link it in the end.

src/eei.h
```diff
@@ -47,6 +50,35 @@ class WasmEngine {
...
+  {
+    if (benchmarkingEnabled)
+      collectBenchmarkingData();
+  }
+
+private:
+  void collectBenchmarkingData();
+
+  using clock = std::chrono::high_resolution_clock;
+  static bool benchmarkingEnabled;
```
axic: I'm suggesting to set this here to `false`.

src/eei.cpp
axic: Maybe there's some confusion here. I do not suggest to move this statement as is, but: https://github.com/ewasm/hera/pull/486/files#r256387031

src/eei.h
poemm: Static member variables must be initialized outside of class.

src/eei.cpp
```diff
@@ -45,6 +46,25 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
...
+  constexpr auto to_us_str = [](clock::duration d) {
+    return std::to_string(std::chrono::duration_cast<std::chrono::microseconds>(d).count());
+  };
+
+  const auto now = clock::now();
+  const auto instantiationDuration = executionStartTime - instantiationStartTime;
+  const auto executionDuration = now - executionStartTime;
+
+  const auto log = "Time [us]: " + to_us_str(instantiationDuration + executionDuration) + " = " +
+                   to_us_str(instantiationDuration) + " + " + to_us_str(executionDuration) + "\n";
```
poemm: Should we print words like "instantiation" and "invocation" here?

poemm:
> I don't think so. Each execution is a new instance of the engine. The only problem is in the output the children come first.

You are correct.

src/eei.h
chfast:
```
/home/chfast/Projects/ewasm/hera/src/eei.h:79:15: error: ISO C++ forbids in-class initialization of non-const static member ‘hera::WasmEngine::benchmarkingEnabled’
   static bool benchmarkingEnabled = false;
               ^~~~~~~~~~~~~~~~~~~
```

src/eei.cpp
```diff
@@ -45,6 +46,25 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
...
+  std::cerr << log;
+  std::ofstream{"hera_benchmarks.log", std::ios::out | std::ios::app} << log;
```
axic: To solve the problem of calls showing the times for children first, perhaps you could pass `msg` to `collectBenchmarkingData` and include some identifiers in the log file, such as `msg.depth`, `msg.sender`, `msg.destination`.

poemm: This is a good idea. But this is unnecessary for all current benchmarks which either (i) execute one function, or (ii) a bencher function calls the function to be benchmarked, possibly multiple times.

chfast:
How about this output format:
```
Time [us]: NNN (instantiation: XXX, execution: YYY)
```

poemm:
Tiny things: it would be easier to parse if `XXX` did not touch `,` and `YYY` did not touch `)`.

axic:
> it would be easier to parse if `XXX` did not touch `,` and `YYY` did not touch `)`.

Why? Actually that should make it easier to parse with regexp.

axic:
Needs a rebase.

poemm:
> Why? Actually that should make it easier to parse with regexp.

Nevermind, I can parse it easily like it is now.

chfast:
Ready?

src/eei.cpp
poemm: For me this is ok, but for others it may be polite to explain the numbers.

chfast:
Changed to 
```
Time [us]: 449 (instantiation: 196, execution: 253)
```

axic:
Tests seem to fail?

chfast:
All green now.

axic:
I still don't like the perhaps non-existent overhead, but don't have the time to "improve" it.

PR #490, opened 2019-02-08
Add call depth level to debugging messages

axic:
Split off #425.

### DIFF
--- a/src/eei.h
+++ b/src/eei.h
@@ -136,6 +136,9 @@ private:
   void eeiRevertOrFinish(bool revert, uint32_t offset, uint32_t size);
 
   // Helpers methods
+  inline std::string depthToString() const {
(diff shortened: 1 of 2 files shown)


axic:
@chfast can you review please?

src/eei.h
```diff
@@ -136,6 +136,9 @@ class EthereumInterface {
   void eeiRevertOrFinish(bool revert, uint32_t offset, uint32_t size);
 
   // Helpers methods
+  inline std::string depthToString() const {
+    return std::string("[") + std::to_string(m_msg.depth) + "]";
```
chfast:
```suggestion
    return "[" + std::to_string(m_msg.depth) + "]";
```

axic: I think it wasn't compiling hence the explicit `string("[")`

axic: Changed.

PR #493, opened 2019-02-08
Update WAVM to 7th Feb 2018

axic:


axic:
@chfast apparently instead of this mess we do in the install scripts, the WAVM author suggests the following: https://github.com/WAVM/WAVM/issues/128

I'm not versed in cmake at all to know if that helps. Can you check?

chfast:
> Can you check?

This will not work in our case unless you add it to Hunter.

axic:
@chfast can you review this? My only requirement for this PR would be that we do not have larger memory leak due to our code than before.

axic:
This is broken after the rebase 😢

PR #495, opened 2019-02-11
Implement getCaller in wavm 

axic:


axic:
@poemm here's `getCaller`.

poemm:
Thank you!

PR #496, opened 2019-02-11
Trigger manual cleanups in WAVM 

axic:


axic:
This doesn't eliminates all leaks, but reduces them significantly.

src/wavm.cpp
```diff
@@ -199,10 +199,18 @@ ExecutionResult WavmEngine::execute(
...
+    ExecutionResult result = internalExecute(context, code, state_code, msg, meterInterfaceGas);
+    // And clean up mess left by this run.
+    Runtime::collectGarbage();
+    return result;
+  } catch (exception const& e) {
+    // And clean up mess left by this run.
+    Runtime::collectGarbage();
+    // We only catch this exception here in order to clean up garbage..
+    // TODO: hopefully WAVM is fixed so that this isn't needed
+    throw e;
```
axic: Should be `throw;` to not loose derived exceptions.

axic:
Do not review/merge yet - have a lot more fixes.

src/wavm.cpp
```diff
@@ -268,21 +275,28 @@ ExecutionResult WavmEngine::internalExecute(
...
-    },
-    [&](Runtime::Exception&& exception) {
-      // FIXME: decide if each of the exception fit into VMTrap/InternalError
-      ensureCondition(false, VMTrap, Runtime::describeException(exception));
-    }
-  );
+    );
+  } catch (exception const& e) {
+    // clean up
+    wavm_host_module::interface.pop();
```
chfast: The `interface` contains raw pointers, are those deleted?

chfast: Is this `wavm_host_module` required by WAVM? It's not thread-save to have a global structure handling execution.

src/wavm.cpp
```diff
@@ -268,21 +275,28 @@ ExecutionResult WavmEngine::internalExecute(
...
-    [&] {
-      try {
-        vector<IR::Value> invokeArgs;
-        Runtime::invokeFunctionChecked(wavm_context, mainFunction, invokeArgs);
-      } catch (EndExecution const&) {
-        // This exception is ignored here because we consider it to be a success.
-        // It is only a clutch for POSIX style exit()
+  // TODO: this is terrible, need to clean up the handling of `interface`
+  try {
+    Runtime::catchRuntimeExceptions(
```
chfast: Does this actually execute anything?

src/wavm.cpp
axic: It is required due to that fact the host functions can be declared via macros only :(

src/wavm.cpp
axic: Yeah, it catches wasm runtime errors (but passes through exceptions in host functions).

src/wavm.cpp
axic: > The `interface` contains raw pointers, are those deleted?

That code was I think suggested by you, can you please check?

chfast: Yes, it is allocated on the stack.

src/wavm.cpp
```diff
@@ -271,21 +279,28 @@ ExecutionResult WavmEngine::internalExecute(
...
-    },
-    [&](Runtime::Exception&& exception) {
-      // FIXME: decide if each of the exception fit into VMTrap/InternalError
-      ensureCondition(false, VMTrap, Runtime::describeException(exception));
-    }
-  );
+    );
+  } catch (exception const&) {
+    // clean up
+    wavm_host_module::interface.pop();
```
axic: This is actually fixing the problem of exceptions not cleaning up between calls.

PR #499, opened 2019-02-12
Some more wavm cleanup

axic:


src/wavm.cpp
```diff
@@ -194,6 +194,35 @@ namespace wavm_host_module {
   };
 } // namespace wavm_host_module
 
+
+void WavmEngine::verifyContract(vector<uint8_t> const& code)
```
axic: @chfast I'm not sure how to do this without code duplication given the mess `GCPointer` is causing. Any ideas?

chfast: What exactly is duplicated?

axic: This entire body below is present in `internalExecute`.

axic:
@chfast it can be reviewed now :wink:

This should conclude the wavm work, everything works now.

src/wavm.cpp
```diff
@@ -393,4 +394,91 @@ ExecutionResult WavmEngine::internalExecute(
...
+
+    ensureCondition(
+      eei_signatures.count(import.exportName),
+      ContractValidationFailure,
+      "Importing invalid EEI method."
+    );
+    IR::FunctionType const& eei_function_type = eei_signatures.at(import.exportName);
+
+    ensureCondition(
+      moduleIR.types.size() >= import.type.index,
```
axic: @chfast is indexing of a vector going from 0 or 1? `moduleIR.types` is a vector.

chfast: From 0.

PR #500, opened 2019-02-12
wavm: Handle WavmEthereumInterface lifetime by RAII

chfast:
Fixes #496.

### DIFF
--- a/src/wavm.cpp
+++ b/src/wavm.cpp
@@ -267,6 +267,15 @@ namespace wavm_host_module {
   };
 } // namespace wavm_host_module
 
+struct WavmInterfaceKeeper {
+  explicit WavmInterfaceKeeper(WavmEthereumInterface& interface)
+  {
+    wavm_host_module::interface.push(&interface);
+  }
+
+  ~WavmInterfaceKeeper() noexcept { wavm_host_module::interface.pop(); }
+};
+
 ExecutionResult WavmEngine::execute(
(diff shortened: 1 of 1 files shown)


src/wavm.cpp
```diff
@@ -215,6 +215,16 @@ ExecutionResult WavmEngine::execute(
   }
 }
 
+struct WavmInterfaceKeeper
```
axic: Can you move this into the `wavm_host_module` namespace? (Just after where `interface` is declared)

src/wavm.cpp
```diff
@@ -280,30 +290,23 @@ ExecutionResult WavmEngine::internalExecute(
...
+  Runtime::catchRuntimeExceptions(
       [&] {
-        try {
+        try
+        {
           vector<IR::Value> invokeArgs;
           Runtime::invokeFunctionChecked(wavm_context, mainFunction, invokeArgs);
-        } catch (EndExecution const&) {
+        }
+        catch (EndExecution const&)
```
axic: Please keep the old style - that is consistent with the rest of the codebase.

axic:
This PR:
```
SUMMARY: AddressSanitizer: 2808 byte(s) leaked in 21 allocation(s).
```
vs

previously:
```
SUMMARY: AddressSanitizer: 2808 byte(s) leaked in 21 allocation(s).
```

So it seems to work.

PR #508, opened 2019-02-26
Simplify code

chfast:


src/hera.cpp
```diff
@@ -131,7 +131,7 @@ pair<evmc_status_code, vector<uint8_t>> callSystemContract(
   if (result.release)
     result.release(&result);
 
-  return make_pair(result.status_code, ret);
+  return {result.status_code, ret};
```
axic: Is this the only instance?

PR #510, opened 2019-03-25
Upgrade wabt to 1.0.10

chfast:


axic:
I tried wabt 1.0.10, but it didnd't compile locally.

PR #511, opened 2019-03-26
Benchmarking geth+evmc with static libs

cdetrio:
Don't merge, for reference only,


cmake/ProjectWabt.cmake
```diff
@@ -22,10 +22,10 @@ ExternalProject_Add(wabt
     CMAKE_ARGS
     -DCMAKE_INSTALL_PREFIX=<INSTALL_DIR>
     -DCMAKE_BUILD_TYPE=Release
-    -DWITH_EXCEPTIONS=ON
+    -DWITH_EXCEPTIONS=OFF
```
axic: Why does this needs to be turned off?

src/binaryen.cpp
```diff
@@ -569,11 +569,15 @@ void BinaryenEngine::verifyContract(wasm::Module & module)
     "Contract export (\"memory\") missing."
   );
 
+/*
+***** in the metering benchmarks, inline-block and inline-super both export the inline useGas functions
```
axic: Why do they export it?

src/hera.cpp
```diff
@@ -92,7 +92,10 @@ struct hera_instance : evmc_instance {
   bool metering = false;
   map<evmc_address, vector<uint8_t>> contract_preload_list;
 
-  hera_instance() noexcept : evmc_instance({EVMC_ABI_VERSION, "hera", hera_get_buildinfo()->project_version, nullptr, nullptr, nullptr, nullptr, nullptr}) {}
+  // hera_instance() noexcept : evmc_instance({EVMC_ABI_VERSION, "hera", hera_get_buildinfo()->project_version, nullptr, nullptr, nullptr, nullptr, nullptr}) {}
+  // fix error with geth+evmc:  /tmp/go-build940146972/b001/runtime.test: symbol lookup error: /tmp/go-build940146972/b001/runtime.test: undefined symbol: hera_get_buildinfo
+  hera_instance() noexcept : evmc_instance({EVMC_ABI_VERSION, "hera", nullptr, nullptr, nullptr, nullptr, nullptr, nullptr}) {}
```
axic: @chfast it seems like `buildinfo` doesn't work in static linking?!

cmake/ProjectWabt.cmake
chfast: It matches wabt default build options.
This breaks some EEI. Long term it would be worth to re-implement EEI without exceptions.

src/hera.cpp
chfast: Hopefully, addressed by https://github.com/ewasm/hera/pull/517.

axic:
Is there something we can salvage from this PR?

src/binaryen.cpp
cdetrio: It was a workaround to make the wast code use the debug name `useGas` for the function (if the inline function wasn't exported, its name would get mangled by the metering injector).

PR #513, opened 2019-04-04
Initial support for runevm contract

s1na:
This PR adds runevm as a system contract. I wanted to be able to test runevm via `testeth`, and made these changes to facilitate that. Not sure if this is something that you'd want merged.

Just now when I wanted to push my branch name conflicted with https://github.com/ewasm/hera/tree/runevm and I found this commit https://github.com/ewasm/hera/commit/866e12c90f644d834eb68836c700ba82c3d90267 (Update: just saw the PR https://github.com/ewasm/hera/pull/375) which would have saved me a day! :)

Three points about the PR:

- I couldn't get `context->host->call` to work, and directly use the `WasmEngine` to call runevm.
- When the runevm instance was being instantiated it would cause out-of-bounds memory access so I added the `module.memory.initial = 20` to avoid that. Not sure how to properly handle this.
- Added a flag `disable-interface-metering`, as runevm currently measures all gas costs itself (again, not sure if this is good or not).

### DIFF
--- a/src/hera.cpp
+++ b/src/hera.cpp
@@ -51,6 +51,14 @@ enum class hera_evm1mode {
   reject,
   fallback,
   evm2wasm_contract,
+  runevm_contract,
+};
+
+const map<string, hera_evm1mode> evm1mode_options {
+  { "reject", hera_evm1mode::reject },
+  { "fallback", hera_evm1mode::fallback },
+  { "evm2wasm", hera_evm1mode::evm2wasm_contract },
+  { "runevm", hera_evm1mode::runevm_contract },
 };
 
 using WasmEngineCreateFn = unique_ptr<WasmEngine>(*)();
@@ -67,25 +75,21 @@ const map<string, WasmEngineCreateFn> wasm_engine_map {
 #endif
 };
 
-const map<string, hera_evm1mode> evm1mode_options {
-  { "reject", hera_evm1mode::reject },
-  { "fallback", hera_evm1mode::fallback },
-  { "evm2wasm", hera_evm1mode::evm2wasm_contract },
-};
-
-struct hera_instance : evmc_instance {
-  unique_ptr<WasmEngine> engine{
+WasmEngineCreateFn wasmEngineCreateFn =
 // This is the order of preference.
 #if HERA_BINARYEN
-    new BinaryenEngine
+    BinaryenEngine::create
 #elif HERA_WABT
-    new WabtEngine
+    WabtEngine::create
 #elif HERA_WAVM
-    new WavmEngine
+    WavmEngine::create
 #else
 #error "No engine requested."
 #endif
-  };
+;
+
+struct hera_instance : evmc_instance {
+  unique_ptr<WasmEngine> engine = wasmEngineCreateFn();
   hera_evm1mode evm1mode = hera_evm1mode::reject;
   bool metering = false;
   map<evmc_address, bytes> contract_preload_list;
@@ -95,6 +99,7 @@ struct hera_instance : evmc_instance {
 
 const evmc_address sentinelAddress = { .bytes = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xa } };
 const evmc_address evm2wasmAddress = { .bytes = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xb } };
+const evmc_address runevmAddress = { .bytes = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0xc } };
 
 // Calls a system contract at @address with input data @input.
 // It is a "staticcall" with sender 000...000 and no value.
@@ -130,6 +135,39 @@ pair<evmc_status_code, bytes> callSystemContract(
     result.release(&result);
 
   return {result.status_code, ret};
+}
+
+pair<evmc_status_code, bytes> locallyExecuteSystemContract(
+  evmc_context* context,
+  evmc_address const& address,
+  int64_t & gas,
+  bytes_view input,
+  bytes_view code,
+  bytes_view state_code
+) {
+  const evmc_message message = {
+    .kind = EVMC_CALL,
+    .flags = EVMC_STATIC,
+    .depth = 0,
+    .gas = gas,
+    .destination = address,
+    .sender = {},
+    .input_data = input.data(),
+    .input_size = input.size(),
+    .value = {},
+    .create2_salt = {},
+  };
+
+  unique_ptr<WasmEngine> engine = wasmEngineCreateFn();
+  // TODO: should we catch exceptions here?
+  ExecutionResult result = engine->execute(context, code, state_code, message, false);
+
+  bytes ret;
+  evmc_status_code status = result.isRevert ? EVMC_REVERT : EVMC_SUCCESS;
+  if (status == EVMC_SUCCESS && result.returnValue.size() > 0)
+    ret = move(result.returnValue);
+
+  return {status, move(ret)};
 }
 
 // Calls the Sentinel contract with input data @input.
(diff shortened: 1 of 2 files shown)


src/hera.cpp
```diff
@@ -73,6 +74,7 @@ const map<string, hera_evm1mode> evm1mode_options {
   { "reject", hera_evm1mode::reject },
   { "fallback", hera_evm1mode::fallback },
   { "evm2wasm", hera_evm1mode::evm2wasm_contract },
+  { "runevm", hera_evm1mode::runevm_contract },
```
axic: Shouldn't need a new entry, it follows the same external API as evm2wasm. Just use it as "evm2wasm".

src/hera.cpp
```diff
@@ -215,7 +262,7 @@ evmc_result hera_execute(
     heraAssert(rev == EVMC_BYZANTIUM, "Only Byzantium supported.");
     heraAssert(msg->gas >= 0, "EVMC supplied negative startgas");
 
-    bool meterInterfaceGas = true;
+    bool meterInterfaceGas = hera->meter_interface;
```
axic: Instead of this, just uncomment these lines:
```
      case hera_evm1mode::evm2wasm_contract:
        run_code = evm2wasm(context, run_code);
        ensureCondition(run_code.size() > 8, ContractValidationFailure, "Transcompiling via evm2wasm failed");
        // TODO: enable this once evm2wasm does metering of interfaces
        // meterInterfaceGas = false;
break;
```

src/binaryen.cpp
```diff
@@ -492,6 +492,7 @@ ExecutionResult BinaryenEngine::execute(
 
   // NOTE: DO NOT use the optimiser here, it will conflict with metering
 
+  module.memory.initial = 20;
```
axic: Why is this needed?

src/hera.cpp
s1na: Can't get the `evm2wasm` mode to work. `callSystemContract` returns empty, as the address has no code associated with it (https://github.com/ethereum/aleth/blob/master/libethereum/Executive.cpp#L342). I expected aleth's evmc to redirect the call to hera again. Here's the command I use:

```
./testeth -t GeneralStateTests/stExample -- --singlenet Byzantium --singletest "add11" --vm ~/dev/ewasm/hera/build/src/libhera.so --evmc evm1mode=evm2wasm --evmc sys:evm2wasm=runevm2.wasm --verbosity 2
```

src/binaryen.cpp
s1na: Runevm's deployer ran out of memory. This was a temporary work-around. Should chisel's deployer maybe run `grow_memory`?

axic: Right, it is due to this: https://github.com/wasmx/wasm-chisel/issues/53

src/hera.cpp
axic: Ah you are correct. We had this issue for it: https://github.com/ethereum/evmc/issues/53

I think we may be better just merging your `runevm` function, but later we could generalise it to have "run locally" vs. "run through the state".

src/binaryen.cpp
s1na: Removed this line.

src/hera.cpp
s1na: Would it be okay to leave this as an evmc flag? it'd be easier to both with interface metering enabled/disabled

src/hera.cpp
```diff
@@ -191,6 +195,49 @@ vector<uint8_t> evm2wasm(evmc_context* context, vector<uint8_t> const& input) {
   return ret;
 }
 
+// Calls the runevm contract.
+// @returns a wasm-based evm interpreter.
+vector<uint8_t> runevm(evmc_context* context, WasmEngine& engine, vector<uint8_t> code) {
```
axic: Can you make a helper `locallyExecuteSystemContract` similar to `callSystemContract`?

src/hera.cpp
```diff
@@ -270,9 +323,6 @@ evmc_result hera_execute(
       );
     }
 
-    heraAssert(hera->engine, "Wasm engine not set.");
-    WasmEngine& engine = *hera->engine;
```
axic: Even though it has a terrible overhead, I think `locallyExecuteSystemContract` should create its own engine instance to avoid any memory/state leaking.

s1na:
Updated, can you have a look @axic?

src/hera.cpp
```diff
@@ -67,34 +68,39 @@ const map<string, WasmEngineCreateFn> wasm_engine_map {
...
-  { "evm2wasm", hera_evm1mode::evm2wasm_contract },
-};
-
-struct hera_instance : evmc_instance {
-  unique_ptr<WasmEngine> engine{
+WasmEngineCreateFn wasmEngineCreateFn =
 // This is the order of preference.
 #if HERA_BINARYEN
-    new BinaryenEngine
+    BinaryenEngine::create
```
axic: @chfast hmm, why did we had the static `create()` if we only ever used the default constructor?

src/hera.cpp
```diff
@@ -247,6 +319,9 @@ evmc_result hera_execute(
         HERA_DEBUG << "Non-WebAssembly input, failure.\n";
         ret.status_code = EVMC_FAILURE;
         return ret;
+      case hera_evm1mode::runevm_contract:
+        run_code = runevm(context, hera->contract_preload_list[runevmAddress]);
+        ensureCondition(run_code.size() > 8, ContractValidationFailure, "Interpreting via runevm failed");
```
axic: Need to set `meterInterfaceGas = false` here.

src/hera.cpp
```diff
@@ -422,10 +497,19 @@ evmc_set_option_result hera_set_option(
     return EVMC_SET_OPTION_INVALID_VALUE;
   }
 
+  if (strcmp(name, "disable-interface-metering") == 0) {
```
axic: I think this isn't needed anymore.

src/hera.cpp
```diff
@@ -132,6 +138,40 @@ pair<evmc_status_code, bytes> callSystemContract(
...
+
+  unique_ptr<WasmEngine> engine = wasmEngineCreateFn();
+  ExecutionResult result = engine->execute(context, code, state_code, message, false);
+
+  bytes ret;
+  evmc_status_code status = result.isRevert ? EVMC_REVERT : EVMC_SUCCESS;
+  if (status == EVMC_SUCCESS && result.returnValue.size() > 0)
+    ret = move(result.returnValue);
+
+  heraAssert(result.gasLeft > 0, "Runevm ran out of gas.");
```
axic: I don't think we need this assertion. The `runevm()` helper below will ensure this.

src/hera.cpp
```diff
@@ -189,6 +229,38 @@ bytes evm2wasm(evmc_context* context, bytes_view input) {
...
+      gas,
+      {},
+      code,
+      code
+  );
+
+  HERA_DEBUG << "runevm done (output " << ret.size() << " bytes) with status=" << status << "\n";
+
+  heraAssert(ret.size() > 0, "Runevm returned empty.");
+  heraAssert(hasWasmPreamble(ret), "Runevm result has no wasm preamble.");
```
axic: Not entirely sure, but I think we should use `ensureCondition` here in both cases.

src/hera.cpp
```diff
@@ -425,7 +505,8 @@ evmc_set_option_result hera_set_option(
   if (strcmp(name, "engine") == 0) {
     auto it = wasm_engine_map.find(value);
     if (it != wasm_engine_map.end()) {
-      hera->engine = it->second();
+      wasmEngineCreateFn = it->second;
+      hera->engine = wasmEngineCreateFn();
```
axic: Sorry, you were correct, `.execute()` is stateless at the moment. Would you mind if I remove the changes regarding the engines and merge this PR?

And then could you create a new PR changing the handling of the engine, because that might be a bug currently, but want to have @chfast's opinion.

s1na: Sure, feel free to modify the PR.

src/hera.cpp
chfast: I don't remember anything about it. If the default constructor works for this case it's good choice.

axic:
@chfast can you please review this PR?

src/hera.cpp
```diff
@@ -132,6 +137,38 @@ pair<evmc_status_code, bytes> callSystemContract(
...
+    .value = {},
+    .create2_salt = {},
+  };
+
+  unique_ptr<WasmEngine> engine = wasmEngineCreateFn();
+  ExecutionResult result = engine->execute(context, code, state_code, message, false);
+
+  bytes ret;
+  evmc_status_code status = result.isRevert ? EVMC_REVERT : EVMC_SUCCESS;
+  if (status == EVMC_SUCCESS && result.returnValue.size() > 0)
```
chfast:
```suggestion
  if (status == EVMC_SUCCESS && !result.returnValue.empty())
```
but the `returnValue` is not really needed.

src/hera.cpp
```diff
@@ -132,6 +137,38 @@ pair<evmc_status_code, bytes> callSystemContract(
...
+    ret = move(result.returnValue);
+
+  return {status, ret};
```
chfast: You have to move `ret` again.

PR #517, opened 2019-04-15
 CMake: Install hera-buildinfo library to allow installing static hera 

chfast:


chfast:
You can verify that https://8200-66198886-gh.circle-artifacts.com/0/package/hera-0.2.3-alpha.0-4%2Bcommit.6af305f5-linux-x86_64.tar.gz contains `hera-buildinfo.a`.

PR #518, opened 2019-04-24
CMake: Hide symbols from wasm engines included

chfast:
For binaryen + wabt:
```
nm -Cg src/libhera.so | wc -l 
readelf -r src/libhera.so | wc -l
```

Before: 9852 10831
After: 3993 6510

### DIFF
--- a/cmake/ProjectWAVM.cmake
+++ b/cmake/ProjectWAVM.cmake
@@ -24,6 +24,7 @@ set(llvmjit_library ${binary_dir}/lib/${CMAKE_STATIC_LIBRARY_PREFIX}LLVMJIT${CMA
 
 set(other_libraries ${platform_library} ${wasm_library} ${ir_library} ${logging_library} ${unwind_library} ${llvmjit_library})
 
+set(flags "-Wno-error -fvisibility=hidden")
 
 ExternalProject_Add(wavm
     PREFIX ${prefix}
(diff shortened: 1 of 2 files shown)


chfast:
I wander why tests on macos are failing...

chfast:
@cdetrio Can you benchmark wabt engine built like this?

axic:
There some exception happening:
```
Totally unknown exception
/Users/distiller/project/test/tools/libtesteth/ImportTest.cpp:768: error: in "GeneralStateTests/stEWASMTests": malformedBytecodeInvalidWasm on Byzantium: Expected another postState hash! expected: 0x0bf1306fe3199b9863a81ed7cb60154a9d94076356eec1fe7654f423999fbcdd actual: 0xf93bc41346c0dad509a9e1ede7e6348b91c9e786999144a44ab5533ecf95698b in Byza
```

Remove that catch statement in `hera.cpp` to see what's wrong.

axic:
```

Executing message in Hera
libc++abi.dylib: terminating with uncaught exception of type wasm::ParseException
unknown location:0: fatal error: in "GeneralStateTests/stEWASMTests": signal: SIGABRT (application abort requested)
/Users/distiller/project/test/tools/libtesteth/TestHelper.cpp:617: last checkpoint
```

Somewhere the parser exception is not caught.

chfast:
> Somewhere the parser exception is not caught in binaryen. `BinaryenEngine::loadModule` is the place it should be usually handled.

Ok, but why it happened now?

axic:
> Ok, but why it happened now?

The only thing I could imagine is that some symbols are runtime resolved and by hiding them that is not possible anymore.

chfast:
Removed the change for Binaryen

PR #519, opened 2019-04-25
Upgrade to EVMC 6.2.0

axic:


axic:
We can do #520 separately from this PR.

PR #521, opened 2019-04-25
Use EVMC HostInterface 

### LINKED ISSUE #520: EVMC: use HostContext in EthereumInterface
See https://github.com/ethereum/evmc/blob/master/include/evmc/evmc.hpp#L160

axic:
Depends on #519. Closes #520.

### DIFF
--- a/src/eei.cpp
+++ b/src/eei.cpp
@@ -102,7 +102,7 @@ void WasmEngine::collectBenchmarkingData()
 
       HERA_DEBUG << "): " << dec;
 
-      evmc_bytes32 result = m_context->host->get_storage(m_context, &m_msg.destination, &path);
+      evmc_bytes32 result = m_host.get_storage(m_msg.destination, path);
 
       if (useHex)
       {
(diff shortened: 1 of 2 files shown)


src/eei.cpp
```diff
@@ -818,9 +818,9 @@ bool exceedsUint128(evmc_uint256be const& value) noexcept
     takeInterfaceGas(GasSchedule::copy * ((int64_t(length) + 31) / 32));
   }
 
-  bool EthereumInterface::enoughSenderBalanceFor(evmc_uint256be const& value) const
+  bool EthereumInterface::enoughSenderBalanceFor(evmc_uint256be const& value)
```
axic: I wonder if we can make assumptions that the host interface will only read and not make changes in certain cases. Perhaps this is a "bug" in Hera.

chfast: I have not considered this before. In practice we can add `const` to every method because it only wraps a pointer to context. Not sure this is what we want. It makes the interface rigid.

chfast:
I believe I fixed this shit. Looks like implicit conversion from `evmc::result` to `evmc_result` is not good idea when `evmc::result` works like smart pointer.

axic:
Ah.. yeah kind of glanced over the fact it wasn't `auto` already.

chfast:
I think I had similar problem before in EVMC, but I still keep the inheritance to allow simple `r.output_size` access. But I think I've found the way now to restrict this conversion: https://godbolt.org/z/tpA1V3.

PR #528, opened 2019-04-26
Add intx as Hunter package

chfast:
It upgrades Hunter and enables it by default. The intx 0.2 is added as the Hunter package.

### DIFF
--- a/cmake/Hunter/init.cmake
+++ b/cmake/Hunter/init.cmake
@@ -1,0 +1,12 @@
+
+if(EVMC_TESTING OR EVMC_TEST_TOOLS)
+    set(HUNTER_CONFIGURATION_TYPES Release
+        CACHE STRING "Build type of the Hunter packages")
+
+    include(HunterGate)
+
+    HunterGate(
+        URL "https://github.com/ruslo/hunter/archive/v0.23.169.tar.gz"
+        SHA1 "60de9f275143ab38df559986c04e3726683e7de7"
+    )
+endif()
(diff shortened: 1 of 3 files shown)


axic:
Will we merge it this way or are we waiting for its inclusion to Hunter? Also any chance release 0.1.1 with the le/be change?

axic:
Are we wainting for https://github.com/ruslo/hunter/pull/1846 to be merged before merging this?

chfast:
Yes

axic:
Why not update cable first?

chfast:
> Why not update cable first

For what?

axic:
It does some of that hunter upgrade & moving stuff, doesn't it?

chfast:
> It does some of that hunter upgrade & moving stuff, doesn't it?

No it doesn't. It only upgraded HunterGate which is the bootstrap script that downloads the Hunter itself. Upgrading it is usually not required. I'm just doing it from time to time for sanity.

cmake/Hunter/init.cmake
```diff
@@ -1,12 +1,10 @@
 
-if(EVMC_TESTING OR EVMC_TEST_TOOLS)
-    set(HUNTER_CONFIGURATION_TYPES Release
-        CACHE STRING "Build type of the Hunter packages")
+set(HUNTER_CONFIGURATION_TYPES Release
+    CACHE STRING "Build type of the Hunter packages")
 
-    include(HunterGate)
+include(HunterGate)
```
axic: How is this link resolved?

Also can you move the whitespace changes to the first commit which introduces the file?

chfast: It's in `cmake/cable` which is added to CMake's module path.

What whitespace changes?

axic: The second commit removes spaces from the file added in the first commit.

chfast: It removes `if()` so the indentation changes.

PR #530, opened 2019-04-26
Implement bignum system library using intx

axic:


src/eei.cpp
```diff
@@ -849,4 +849,35 @@ void WasmEngine::collectBenchmarkingData()
...
+    storeBignum256(loadBignum256(aOffset) * loadBignum256(bOffset), retOffset);
+  }
+
+  void EthereumInterface::umulmod256(uint32_t aOffset, uint32_t bOffset, uint32_t modOffset, uint32_t retOffset)
+  {
+    using intx::uint512;
+    auto a = loadBignum256(aOffset);
+    auto b = loadBignum256(bOffset);
+    auto mod = loadBignum256(modOffset);
+    auto ret = mod != 0 ? ((uint512{a} * uint512{b}) % uint512{mod}).lo : 0;
```
axic: This above logic is from evmone: https://github.com/chfast/evmone/blob/master/lib/evmone/execution.cpp#L121

src/eei.cpp
```diff
@@ -849,4 +849,35 @@ void WasmEngine::collectBenchmarkingData()
     }
     return ret;
   }
+
+  intx::uint256 EthereumInterface::loadBignum256(uint32_t srcOffset)
+  {
+    uint8_t data[32];
+    loadMemory(srcOffset, data, 32);
+    // FIXME: change this to little endian?
```
axic: intx only supports big endian right now: https://github.com/chfast/intx/issues/78

axic:
Need to add support to binaryen.

PR #533, opened 2019-04-29
Clean up includes

axic:
Fixes some includes after #484.


axic:
> The `#include <memory>` is missing.

I guess macOS is different again, because it compiled her.

chfast:
> I guess macOS is different again, because it compiled her.

This happens all the time both ways because of libstdc++ and libc++ differences.

PR #536, opened 2019-04-30
Attempt to fix lgtm configuration

axic:


axic:
Need to merge this because apparently LGTM doesn't get the config from PRs :(

PR #540, opened 2019-05-17
wabt: print errors during contract validation and execution

jwasinger:


src/wabt.cpp
```diff
@@ -636,6 +636,11 @@ ExecutionResult WabtEngine::execute(
   );
 
   ensureCondition(Succeeded(loadResult) && module, ContractValidationFailure, "Module failed to load.");
+
+  for (auto it = errors.begin(); it != errors.end(); ++it) {
```
axic: Can you wrap these under `#if HERA_DEBUGGING`?

Also I don't think this will work at all, because the `ensureCondition(Succeeded(loadResult)` will abort execution before this for loop.

jwasinger: Oh. true.

PR #545, opened 2019-05-22
Update to cable 0.4.0

axic:


axic:
@chfast is there any other hunter improvements we can make here (since #528 was merged)?

chfast:
There are not improvements expected. Rather build issues as Cable 0.4 enables more warnings by default.

axic:
To fix this PR need to update EVMC to remove shadowing. Will wait for the EVMC 6.3 release.

axic:
This depends on an EVMC release + update, which fixed the shadowing issues.

PR #546, opened 2019-05-22
Run codecov on all engines

axic:


axic:
@chfast can you have a look why codecov is failing?

chfast:
The compiler went OOM.

cmake/ProjectWAVM.cmake
```diff
@@ -26,6 +26,10 @@ set(other_libraries ${platform_library} ${wasm_library} ${ir_library} ${logging_
 
 set(flags "-Wno-error -fvisibility=hidden")
 
+if(CMAKE_GENERATOR STREQUAL Ninja)
+    set(build_command BUILD_COMMAND ninja -j1)
```
axic: Any way to use the concurrency flag from cmake (if there's one)?

axic:
@chfast would it be a lot of work splitting up coverage into:
- running all engines (this will not allow us to see if the eei is covered properly)
- running only wabt to see if it covers the entire eei properly

cmake/ProjectWAVM.cmake
chfast: Somehow this flag is not passed down to external projects for ninja. We can switch to make to make is simple.

chfast:
> @chfast would it be a lot of work splitting up coverage into:
> 
>     * running all engines (this will not allow us to see if the eei is covered properly)
> 
>     * running only wabt to see if it covers the entire eei properly

A bit. On codecov you can tag different data uploads, but codecov is not able to filter these in web UI. Maybe coveralls can do it.

axic:
@chfast thanks for fixing this!

PR #549, opened 2019-07-17
cmake: Use evmc-vmtester helper

chfast:


circle.yml
```diff
@@ -163,8 +163,9 @@ defaults:
   evmc-test: &evmc-test
     run:
       name: "Run evmc tests"
+      working_directory: ~/build
       command: |
-        ~/build/evmc/test/evmc-vmtester ~/build/src/libhera.so
+        ctest
```
axic: Where is this name coming from?

chfast: https://cmake.org/cmake/help/latest/manual/ctest.1.html

test/CMakeLists.txt
```diff
@@ -1,3 +1,9 @@
+include(../evmc/cmake/EVMC.cmake)
+
+if(TARGET evmc::evmc-vmtester)
+    evmc_add_vm_test(NAME evmc-check TARGET hera)
```
axic: @chfast Is this statically linking hera?

PR #558, opened 2020-06-09
Use C++ evmc wrappers for everything (including uint256be)

axic:


src/hera.cpp
```diff
@@ -24,6 +24,7 @@
 #include <memory>
 
 #include <evmc/evmc.h>
+#include <evmc/evmc.hpp>
```
chfast, line 27: You don't need to include `evmc.h` when `evmc.hpp` is included.

PR #565, opened 2021-05-12
Upgrade EVMC to version 8.0.0

axic:


circle.yml
```diff
@@ -212,11 +212,11 @@ jobs:
       - *install-aleth
       - *add-package-to-workspace
       - *fetch-tests
-      - *test-binaryen
-      - *test-wabt
-      - *test-wavm
+#      - *test-binaryen
```
chfast, line 215: You should add comment why this is disabled.

chfast (approved):

circle.yml
axic: You already disabled evmc-test earlier and no comment is to be found 😉 

However the reason can be found using git blame + git log as it is in the commit message.

